PR #9, opened 2015-09-16
Update to prose.io config

jamesonzimmer:
Promised to check this out a long time ago! This will make it so folks can edit their posts with a WYSIWYG editor if they so desire.


daattali:
Thanks @jamesonzimmer !

jamesonzimmer:
No problem! Theme looks great and I love the mission.

PR #17, opened 2015-11-21
Vagrant support

XNerv:
Add vagrant support for easy local development


daattali:
Thanks for the good quality commit and even adding good documentation!

PR #27, opened 2016-01-07
Fix GPG key error in scripted installation of RVM

XNerv:
This problem appeared suddenly. For more info see https://github.com/rvm/rvm/issues/3110

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -4,7 +4,8 @@
 $script_provision = <<SCRIPT
   echo Provisioning...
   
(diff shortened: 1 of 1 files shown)


daattali:
I'll have to trust you on this one and accept this PR without actually testing it since I don't use this.  Thanks for following up with this.

XNerv:
I had created a new vagrant instance of my blog, which worked a few days ago, and this error thrown to me. This PR fix this. 

Thank you for the merge.

PR #42, opened 2016-03-03
bah

tudr:
back to normal?

[Review on Reviewable](https://reviewable.io/reviews/daattali/beautiful-jekyll/42)


daattali:
You need to make these changes on your copy of the repository, not mine

tudr:
Yup, am aware. Have had a lot of trouble getting my local copy to it's previous
state, how would you approach it?

On Thu, Mar 3, 2016 at 5:23 PM Dean Attali <notifications@github.com> wrote:
You need to make these changes on your copy of the repository, not mine

daattali:
I'm not sure what you're trying to do.  If you want to get your repo back to a previous state, you can't do it from GitHub, you'll have to download git and use git commands. If you're not comfortable using git and can't figure out how to do that, then there isn't really a simple way. The best option would be to delete the repository entirely and re-fork beautiful jekyll to start again

tudr:
Awesome, sounds like we think alike. I spent a bunch of time trying to figure it
out through git, but ended up starting fresh :)
Thanks, again.

On Thu, Mar 3, 2016 at 6:29 PM Dean Attali <notifications@github.com> wrote:
I'm not sure what you're trying to do. If you want to get your repo back to a
previous state, you can't do it from GitHub, you'll have to download git and use
git commands. If you're not comfortable using git and can't figure out how to do
that, then there isn't really a simple way. The best option would be to delete
the repository entirely and re-fork beautiful jekyll to start again

PR #46, opened 2016-03-29
Update _config.yml

tonomoshia:
added my information


daattali:
You need to edit this file on your copy of the project, not mine

PR #55, opened 2016-05-14
Upgrade gems including Jekyll

kendaleiv:
This upgrades Jekyll from `2.x` to `3.x`.

In order to use triple backtick code blocks I needed to adjust the `_config.yml`, which is not included in this PR. If you'd like that incorporated prior to merge, let me know and I'll add the following in another commit to `_config.yml`:

```
kramdown:
  input: GFM
```


daattali:
Thanks!

If we add the input:gfm to the config, I'm not sure if there are any problems it would introduce into existing pages? Or does it only add support for some github-flavoured syntax, without causing any conflicts?

PR #56, opened 2016-05-15
Support triple backticks

kendaleiv:
As mentioned in #55, this makes triple backticks work as expected.

Left is expected existing, right is expected new:

image

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -78,6 +78,9 @@ highlighter: rouge
 permalink: /:year-:month-:day-:title/
 paginate: 5
 
+kramdown:
+  input: GFM
+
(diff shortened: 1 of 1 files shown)


daattali:
I understand this will add the ability to do code chunks when using jekyll locally. Do you know if this will affect in any way GitHub Pages?

kendaleiv:
It seems to work with GitHub pages on my personal blog. See http://kendaleiv.com/typescript-wheres-my-csharp-style-object-initializers/ for an example.

I don't know if it *fixes* a problem using GitHub pages. But, it doesn't seem to break it in my experience.

PR #62, opened 2016-06-13
Upgrade Gemfile.lock (so Jekyll >= 3.1.4) to avoid bug on page/layout metadata

erikmd:
Hi,

Thanks for sharing your nice Jekyll theme!

I recently tested both `master` and `gh-pages` branches of https://github.com/daattali/beautiful-jekyll
by locally doing `bundle exec jekyll serve` as suggested on https://help.github.com/articles/setting-up-your-github-pages-site-locally-with-jekyll/#step-4-build-your-local-jekyll-site and I noticed that only the `gh-pages` branch worked out-of-the-box, while the css rendering was broken for `master` branch.

After investigating a bit, I noticed that the issue was exactly that of https://github.com/jekyll/jekyll/issues/4439 and that the `Gemfile.lock` file (only provided in the `master` branch) was forcing an older version of Jekyll (namely 3.0.5 < 3.1.4).

So to fix this I just ran the following commands to regenerate the `Gemfile.lock`:
```
rm Gemfile.lock
gem install --user-install bundler    # because I install gems locally, not system-wide
bundle install --path ~/.gem          # idem
bundle exec jekyll serve
```
and after pushing branch `master`, the GitHub Pages was indeed OK; hence this pull request.

Actually another solution could be considered: we could just as well remove the `Gemfile.lock` file... but according to these two blog articles it might seem a good idea to keep this file:
http://mattallan.org/2016/github-pages-best-practices/
http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/

Kind regards.


daattali:
Looks good, I still never tried installing locally to keep it simple so I'll trust that these new versions are all correct. Thanks!

PR #63, opened 2016-06-20
Change order of post excerpt processing

lrdodge:
Since the truncate happens first, parts of HTML tags are cut in ways which will not be caught by the strip_html and xml_escape string filters. Switching the order of the filters fixes this issue. An example of this problem can be seen in the Test Mardown post.


daattali:
good catch, thanks!

PR #73, opened 2016-07-27
add tags

epwalsh:


daattali:
Thank you @epwalsh , I appreciate the time to make a PR. I tend to prefer to keep Beautiful Jekyll more focused on working only within GitHub Pages without requiring local installation. This will only work if I use jekyll locally, correct?

epwalsh:
@daattali yea.. I hadn't thought of that since I deploy my site on a DigitalOcean server. No worries!

daattali:
Hmm I might re-consider and add this because I do know there have been a few people in the past who've asked about this.  Do you think you could make a PR again, but without altering the site at all for people who don't have this turned on? I noticed that your site changed the index page a bit which I would want to avoid. Maybe there can be a parameter in the config for whether or not to show tags on the index page. And the tags that are currently shown on a post's page could become hyperlinks if the tags page param is turned on.

I realize that's quite a few things I'm asking, so no pressure, but if you do add all that in (and the corresponding documentation in the README :p ) then I'll merge!  Thanks regardless

epwalsh:
OK sure, no problem. If the option "show-tags" is false, do you still want the tags to show up just as text the blog posts? Or not at all?

daattali:
Ok, I take back what I said earlier about not showing tags on the front page. Right now, none of the sample posts in Beautiful Jekyll have tags, and I think most people who fork this repo don't use tags. But if someone does add tags, then it's fine to show them on the index page, and also of course on the post's page.  And if it's not a local jekyll installation that supports the tags feature, then it's fine to show the plain text tags in both pages.  I haven't slept last night, so I'm sorry if I'm not very clear!

epwalsh:
Lol no worries. I just submitted a new PR

PR #74, opened 2016-07-27
Add tag options

epwalsh:
Tags will be displayed as plain text by default. If user is not deploying on GitHub pages, they can set `link-tags: true` in `_config.yml` to tell Jekyll to automatically create a new page for each tag that lists all posts that have that tag. I've added a brief explanation under "Advanced featrues" in the README.

### DIFF
--- a/_layouts/tag_index.html
+++ b/_layouts/tag_index.html
@@ -1,0 +1,42 @@
+---
+layout: default
+show-avatar: false
+---
+
+<h2 class="post_title">{{page.title}}:</h2>
+
+<div class="posts-list">
+  {% for post in site.posts %}
+  {% for tag in post.tags %}
+  {% if tag == page.tag %}
+  <article class="post-preview">
+    <a href="{{ post.url | prepend: site.baseurl }}">
+      <h3 class="post-title">{{ post.title }}</h3>
+
+      {% if post.subtitle %}
+      <h4 class="post-subtitle">
+        {{ post.subtitle }}
+      </h4>
+      {% endif %}
+    </a>
+
+    <p class="post-meta">
+    Posted on {{ post.date | date: "%B %-d, %Y" }}
+    </p>
+
+    <div class="post-entry">
+      {{ post.content | strip_html | xml_escape | truncatewords: 50 }}
+      <a href="{{ post.url | prepend: site.baseurl }}" class="post-read-more">[Read&nbsp;More]</a>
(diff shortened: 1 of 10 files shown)


index.html
```diff
@@ -26,6 +26,19 @@ <h3 class="post-subtitle">
 	  <a href="{{ post.url | prepend: site.baseurl }}" class="post-read-more">[Read&nbsp;More]</a>
     </div>
 
+    {% if post.tags.size > 0 %}
+    <div class="blog-tags">
+      Tags: 
+      {% if site.link-tags %}
+      {% for tag in post.tags %}
+        <a href="/tag/{{ tag }}">{{ tag }}</a>
```
daattali: I think it's safer to prepend site.baseurl here so that it'll work with GitHub project sites as well

css/main.css
```diff
@@ -331,6 +331,34 @@ footer .theme-by {
...
 }
 
+/* --- Tags --- */
+
+.blog-tags {
+  font-family: 'Open Sans', 'Helvetica Neue', Helvetica, Arial, sans-serif;
+  color: #999;
+  font-size: 15px;
+  padding-top: 5px;
+  /* margin-bottom: 5px; */
```
daattali: could you please remove this rule (and the next documented one for color:#999)

_layouts/tag_index.html
```diff
@@ -0,0 +1,42 @@
+---
+layout: default
+show-avatar: false
```
daattali: in order to have a more consistent experience, I'd prefer to not set "show-avatar" to false by default

css/main.css
epwalsh: Do you just mean the ones that are commented out?

daattali:
Thanks for acting so quick and making such a high quality PR, even added tags to sample posts. Great stuff. I'll merge after the small nitpicky things I mentioned are fixed

css/main.css
daattali: Yeah, just the commented out ones

index.html
epwalsh: So line 34 should be `<a href="{{ site.baseurl }}/tag/{{ tag }}">...` right?

daattali: Correct :+1:

daattali:
Thank you!

epwalsh:
Glad I could help out!

PR #75, opened 2016-07-28
open external links in a separate window

amar-sanakal:
Hello Dean,

Firstly, thank you for this beautiful theme :+1:  I've made this minor change to open external links in a separate window. If you like this feature, you can merge this pull request.

Regards,
Amar


daattali:
Hi Amar, opening links in new tabs is generally considered bad practice and not user friendly,  so I'd prefer not to add this functionality.  Thanks though!

amar-sanakal:
Hello Dean,

Thanks for the feedback. It certainly made me look up UI design articles and I'm now a much wiser man :) In fact, I'll get rid of that change from my website as well.

Regards,
Amar

PR #80, opened 2016-08-01
Add social share buttons for posts

rtlee9:
I added social share buttons to my fork and thought it could be of interest to you and the Beautiful Jekyll community. I've attached a screenshot below, and you can find a live demo [here](http://eightportions.com/2016-07-14-NYC-yellow-cabs-neighborhoods/#fn:6) as well. I've added a few lines to the config file so users can select which (if any) social channels they want to include.

Please let me know if you have any edits - I'm still learning HTML, CSS and the works so I won't be insulted.

Known issue: when buttons overflow (screen width < ~435px with four share buttons activated), there is no spacing in between the first and second rows of buttons. Unsure how to solve this at present -- still in learning-mode, per above.

screen shot 2016-07-31 at 10 12 01 pm


daattali:
Thank you Ryan, I appreciate the contribution.  I'm a bit swamped right now
but I'll take a look sometime this week

rtlee9:
Sure thing. When you do get around to it note that I just incorporated a fix for the overflow issue (added margin-bottom to the button class) into the same commit.

daattali:
Great commits, really nice code. Thank you.  But I noticed you're using a `<share>` tag - is that a valid HTML tag? I've never seen it before, you're sure it's correct?

rtlee9:
Good find - I've replaced the `<share>` tag with a `<section>` tag and rebased onto your master. There's a long story behind the `<share>` tag loosely related to [this](http://stackoverflow.com/questions/20353613/why-does-css-work-with-fake-elements) question, which I won't go into...

daattali:
Yeah you could add any tag you'd like but I prefer to stick with valid
standard tags. I like that you even took care of the case when no tags are
added and looked for that in a variable, good job

PR #82, opened 2016-08-05
Change footer position to relative

lrdodge:
Fixed position was causing the footer to cover content at the bottom of the page.


daattali:
Can you show me a page with the problem?

daattali:
I see the problem. There is CSS to fix this, I recently accidentally took out the CSS class for it. I'll add it back in.  Thank you for pointing this out!

PR #89, opened 2016-08-22
Add keyword parameter

mominul:
Hi,
I don't know if this patch is eligible for merge, but I am using this feature in my own copy.

But I want to thank you for making this wonderful theme!

Thanks again!


daattali:
Thank you @mominul , but I think this is identical to what the `tags` param does at the moment.  I didn't document it because it's not very useful if you have a github pages site, but if you develop locally then you can have the tags be links

mominul:
Thanks for your clarification!

PR #101, opened 2016-09-15
use post.excerpt instead of post.content

lbsx:
In index.html, there is a line for post excerpts:
```
      {{ post.content | strip_html | xml_escape | truncatewords: 50 }}
```
But Chinese and Japanese don't contain spaces in articles, so it's not possible to truncate 50 words from that kind of posts.

image

I made a test shown in the image. If there aren't enough spaces in Chinese posts, the whole article is used as the excerpt.
 
Would you consider using [post.excerpt](https://jekyllrb.com/docs/posts/#post-excerpts) instead?
```
      {{ post.excerpt | strip_html | xml_escape | truncatewords: 50 }}
```

This is how it looks.

image

### DIFF
--- a/index.html
+++ b/index.html
@@ -22,7 +22,7 @@ subtitle: This is where I will tell my friends way too much about me
     </p>
 
     <div class="post-entry">
-      {{ post.content | strip_html | xml_escape | truncatewords: 50 }}
+      {{ post.excerpt | strip_html | xml_escape | truncatewords: 50 }}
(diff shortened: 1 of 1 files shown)


daattali:
Hi @lbsx thanks for the PR! I see your point and I agree that's not ideal.  I'm not perfectly happy with the proposed solution either though, because if a post begins with one very short sentence and then a main paragraph, it will only show those few words in the beginning.  But it's a less major problem, so I'll accept :)

PR #105, opened 2016-09-26
add fb app ID, improve social sharing descriptions

vietanhdev:
I added FB app ID option, edited some social sharing details so that it can be better.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -19,14 +19,6 @@ title: My website
 # Short description of your site
 description: A virtual proof that name is awesome!
 
-# --- Misc --- #
-
-# Fill in your Google Analytics ID to track your website using GA
-# google_analytics: "UA-78252451-1"
-
-# Facebook App ID
-# fb_app_id: "1794997907410000"
-
 # --- Navigation bar options --- #
 
 # List of links in the navigation bar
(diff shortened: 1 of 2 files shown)


_includes/head.html
```diff
@@ -49,32 +49,69 @@
...
   <meta property="og:url" content="{{ site.url }}{{ page.url }}" />
+  <link rel="canonical" href="{{ site.url }}{{ page.url }}" />
   {% else %}
   <meta property="og:url" content="{{ site.url }}{{ page.url | remove: '/index.html' | remove: '.html' }}" />
+  <link rel="canonical" href="{{ site.url }}{{ page.url | remove: '/index.html' | remove: '.html' }}" />
   {% endif %}
+
   {% if page.share-img %}
   <meta property="og:image" content="{{ page.share-img }}" />
-  {% elsif site.avatar %}
```
daattali: Why did you remove this line? If there's no avatar, then we shouldn't serve {{site.url}} as the image

daattali: Great commit otherwise

vietanhdev: Sorry I've fixed it. Can you review it?

_config.yml
```diff
@@ -19,6 +19,14 @@ title: My website
 # Short description of your site
 description: A virtual proof that name is awesome!
 
+# --- Misc --- #
+
+# Fill in your Google Analytics ID to track your website using GA
+# google_analytics: "UA-78252451-1"
```
daattali: could you remove this google_analytics line (since it already exists below)

_config.yml
```diff
@@ -19,6 +19,14 @@ title: My website
...
+
+# Facebook App ID
+# fb_app_id: "1794997907410000"
```
daattali: could you move this line below to the Misc section, and remove your app ID please?

vietanhdev: Okey. I've done this.

daattali:
thank you!

PR #106, opened 2016-09-29
Open social links in a new tab

anuditverma:
Just added target="_blank" to each anchor tags in order to open the social links in a new tab. BTW awesome work, I'm currently moving from the other most popular CMS to Jekyll, soon to have the site up and running with beautiful-jekyll.

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -5,7 +5,7 @@
         <ul class="list-inline text-center footer-links">
           {% if site.author.facebook and site.footer-links-active.facebook %}
           <li>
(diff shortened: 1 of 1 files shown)


daattali:
Hi @anuditverma thanks for contributing! But I purposely didn't make the links open in new tabs, and I'd like to keep it that way. Yes, it may be "bad for business" to make the user leave your site, but it's considered very poor user experience to open links in new tabs and people usually agree it's bad practice to do that unless for really good reasons. So I prefer to keep links open in current tab. Thanks though!

PR #110, opened 2016-10-04
Add check to only show readme link when there is additional post cont…

lrdodge:
Add check to only show read me link when there is additional post content not shown on index page. Handles truncate words problems with Chinese and Japanese characters by checking for a manual break using the site excerpt separator.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -81,6 +81,8 @@ share-links-active:
 url-pretty: "MyWebsite.com"  # eg. "deanattali.com/beautiful-jekyll"
 
 # --- Misc --- #
+# Manually break the site excerpt
+excerpt_separator: <!--read more-->
 
 # Fill in your Disqus shortname (NOT the userid) if you want to support Disqus comments
 #disqus: ""
--- a/index.html
+++ b/index.html
@@ -22,8 +22,19 @@ subtitle: This is where I will tell my friends way too much about me
     </p>
 
     <div class="post-entry">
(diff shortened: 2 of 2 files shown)


daattali:
great quality PR, thanks!

daattali:
@lrdodge actually, I notice now that there's a problem with this update.  By default, jekyll treats a pargraph as an excerpt, which is great because you'll never get more than one pargaraph in the excerpt.  But now that you defined a new excerpt_separator, jekyll no longer considers the paragraph to be an excerpt, so now multiple paragraphs will be bunched together (if the first one is < 50 words).

I'm going to revert this commit. If you can make you commit again but while preserving paragraphs, that'd be great, because I do like this

lrdodge:
Ignoring paragraphs as breaks for the excerpt was intentional. The idea was to keep your original intent truncating the content to fifty words instead of truncating the excerpt. You mentioned in #101 the issue of short paragraphs being an issue with using the excerpt. Manual breaking was the compromise between not having short excerpts on the index page and handling Chinese/Japanese characters. This was the only solution I could come up with which did both right out of the box.

The funny thing is, my original prototype I was using on my site didn't use the excerpt separator and is exactly how you changed it back to. I only added the separator to the pull request because I thought that'd be how you want it. So much for my mind reading abilities. 😄

lrdodge:
By switching back to using excerpt instead of content the read me link does not show on excerpts under fifty words, regardless if there is more content. I should have time to create a new pull request tomorrow to fix the issue.

PR #111, opened 2016-10-06
Fix show/hide of Read Me link on index page

lrdodge:
Uses the excerpt only to a maximum number of words on the index page. Moved out maximum number of excerpt words value to configuration.


daattali:
Thanks!

PR #124, opened 2016-11-09
add TOC in the navigation

anbuchelva:
adding the table of contents in the navigation bar

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -23,6 +23,7 @@ description: A virtual proof that name is awesome!
 
 # List of links in the navigation bar
 navbar-links:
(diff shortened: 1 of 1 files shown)


daattali:
You need to do this on *your* copy of the project, not here

anbuchelva:
thought it will be useful for everyone.. no probs

daattali:
Oh you actually did mean to make a PR (99% of the PRs I get are people who think they're changing their own copy)

The index.html page has a "table of contents" so I don't think it's needed in the nav, plus "TOC" is a term that the average person doesn't understand what it means

PR #128, opened 2016-11-12
Windows fix for Gemfile, removes execjs and therubyracer from dependencies

adrianharabula:
`bundle install` didn't work on windows because therubyracer could not compile.

So I removed it.

Now you can run `bundle install` on windows.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,4 @@
 source 'https://rubygems.org'
 
-gem 'execjs'
-gem 'therubyracer'
 gem 'github-pages'
 gem 'jekyll-paginate'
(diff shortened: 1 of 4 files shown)


daattali:
I've never tried this so I'm not sure what you removed - can you let me know what therubyracer does and if you're 100% sure it's safe to remove, even for non-windows?

adrianharabula:
I don't know what execjs and therubyracer packages are doing, but I could not make them compile on windows.

On official [github pages docs](https://help.github.com/articles/setting-up-your-github-pages-site-locally-with-jekyll/#step-2-install-jekyll-using-bundler) the Gemfile looks like this:
```
source 'https://rubygems.org'
gem 'github-pages', group: :jekyll_plugins
```

So no execjs and therubyracer.

I tested the current configuration on ubuntu 16.10 with ruby 2.3.0, Windows with ruby 2.3.1p112 and OS X with ruby 2.2 and the blog could be loaded properly without execjs and therubyracer.

daattali:
thanks for testing

PR #134, opened 2016-12-05
Update github-pages gem

stephentuso:
I am serving this locally with `jekyll serve` (after running `bundle install`). However, see the [GitHub pages gem readme](https://github.com/github/pages-gem), the version has to be explicitly stated due to a bundler issue, otherwise a very old version will be used.

This PR just updates the github-pages gem (and its dependencies).

Side note: `bundle install` had an issue installing one of the dependencies, `nokogiri`, but with some help from [stackoverflow](http://stackoverflow.com/questions/26878263/libxml2-missing-mac-os-x-10-10) I got it installed with 
```
sudo env ARCHFLAGS="-arch x86_64" gem install nokogiri:1.8.4.1 -- --with-xml=/usr/local/Cellar/libxml2/2.9.2/
```


daattali:
Thanks!

PR #143, opened 2016-12-24
Allow dynamic images on each blog post

bbritten:
This PR would allow users to add `image: /img/some_picture.jpeg` to their post YAML, which will result in the specified image showing up next to their post excerpt, and also where the site avatar would be when they read the post. All default activity should still be in place if an image is not included. Only thing to note that is that if an image is used, it will override the site avatar for that post.


daattali:
Awesome PR, thanks!  I'm usually very picky about code I accept, but this is great. Two small comments:
- could you take care of the responsiveness css? I tried loading your website on my phone and the image was covering part of the date and was way too big
- could you add a short blurb about this parameter is the README, in the section where I have a table with all the params?

bbritten:
Absolutely. Responsive CSS isn't my forte, so give me a little bit to get it working properly (should be easy enough to copy the responsive CSS you have implemented) and I'll recommit.

daattali:
Nothing fancy required for the css - just add a position and a size that is more appropriate for smaller screens (say, <400 or <500 px)

bbritten:
Please let me know what you think of image placements on small screens. If you have a specific placement/size in mind, I'm more than happy to implement those exact specifications.

daattali:
Looks good, thanks!

bbritten:
I'm glad you like it! Have a good night and happy holidays!

daattali:
Merry Christmas :tada:

daattali:
@bbritten just FYI - not even one day later and someone already  proved my point that making this look good on mobile was important :) https://github.com/daattali/beautiful-jekyll/pull/144

bbritten:
Also answers my question about how much time I should dedicate to learning responsive CSS.

PRElias:
Is there a way to use this without override the site avatar? thanks

PR #144, opened 2016-12-25
Fixed: Post image hides posts' date

anuditverma:
Hi, I was implementing these changes on my site, I noticed that post image was hiding/covering the post's dates (mostly the year) when viewed on small screen devices.
-13px seems to work (not getting much drifted away when viewing on desktop layout), I hope it's fine, I guess you have something else in mind to fix this.
Thanks.

### DIFF
--- a/css/main.css
+++ b/css/main.css
@@ -351,7 +351,7 @@ footer .theme-by {
   float: right;
   height: 192px;
   width: 192px;
-  margin-top: -35px;
+  margin-top: -13px;
   filter: grayscale(90%);
 }
 .post-image:hover {


daattali:
Did you manually copy yesterday's updates to your site? Because there is CSS that makes sure the style is different on mobile screens but your site doesn't seem to have it

https://github.com/daattali/beautiful-jekyll/blob/master/css/main.css#L407-L419

anuditverma:
@daattali Oh I see now, Yes I did some updates manually, I think I might have missed this one but anyways thanks for looking into this.

PR #145, opened 2016-12-25
Added footer-links-active support to footer.html

sasivarnan:
Taken `footer.html` file content from the `master` branch and added it in `footer.html` of `gem` branch.

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -3,7 +3,7 @@
     <div class="row">
       <div class="col-lg-8 col-lg-offset-2 col-md-10 col-md-offset-1">
         <ul class="list-inline text-center footer-links">
-          {% if site.author.facebook %}
+          {% if site.author.facebook and site.footer-links-active.facebook %}
           <li>
             <a href="https://www.facebook.com/{{ site.author.facebook }}" title="Facebook">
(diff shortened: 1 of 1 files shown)


daattali:
@sasivarnan thank you for this PR, but I actually left out `footer-links-active` on purpose. I think it was a mistake on the master branch because it is mostly redundant and just complicates the footer. If there's a link you want to show, you include it. If not, delete it or comment it out. I don't think there needs to be two parameters for each circle. When I first developed beautiful-jekyll there were only 5 circles so it wasn't too much, but now there are so many that I don't think it makes sense to have two parameters for each

@razu55 - were you asked to review today's 2 PRs?

sasivarnan:
@daattali I felt the complication too. The reason why I raised this request is, the footer template in `gem` doesn't had `RSS` feed icon, which was available along with `footer-links-active` in master branch.

daattali:
I understand. That really is the only disadvantage of the new system, and I did think about it but I assumed not enough people use RSS to make it important to support. But since it looks like someone does use it, I'll release a new gem later and support RSS (as a separate param though, not with the footer-links-active)

sasivarnan:
@daattali. Thanks a lot. 😄

PR #154, opened 2017-01-05
Add options to configure HTML document title

tony-ho:
Add `use-site-title` YAML front matter parameter.
If you want to use the site title rather than page title as HTML document title (ie. browser tab title), use `use-site-title: true`. When set, the document title will take the format `Site Title - Site Description` (eg. `My website - A virtual proof that name is awesome!`). Otherwise, it will default to `Page Title - Site Title` (eg. `About me - My website`).

Add `title-separator` site variable.
If you want to set the character used to separate site title and description in HTML document title and RSS feed title. (eg. `title-separator: "|"` will result in title `My website | A virtual proof that name is awesome!`)


_includes/head.html
```diff
@@ -3,15 +3,15 @@
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
   <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0">
 
-  <title>{% if page.title %}{{ page.title }}{% else %}{{ site.title }}{% endif %}</title>
+  <title>{% if page.use-site-title %}{{ site.title }} {{ site.title-separator }} {{ site.description }}{% else %}{{ page.title }} {{ site.title-separator }} {{ site.title }}{% endif %}</title>
```
daattali: In the case of not using `use-site-title`, I'd want to keep the old behaviour (use page.title if it exists, or site.title otherwise)

daattali:
Thanks for your PR. Could you please just address the one comment I had

tony-ho:
Addressed PR feedback. To completely retain old behaviour, can also revert the change to index.html.

daattali:
The change on the index page is good. Thanks!

PR #164, opened 2017-01-27
Add Subresource Integrity (SRI) support

tony-ho:
Modify `ext-js` and `ext-css` YAML front matter parameters to take
`href` and `sri` values. If `sri` is set, add to integrity attribute on
script/link element loading external JavaScript/CSS.

Existing behaviour is maintained if `href`/`sri` are not used.

For more information on SRI, see:
https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity


daattali:
This is a very high quality PR, thank you!

PR #180, opened 2017-03-14
Added my site on the list of personal sites

SharePointOscar:
Added my site under Personal Websites section.


daattali:
you should also add it to the featured users page (in the gh-pages branch)

PR #242, opened 2017-08-31
adds background pattern

OCram85:
Adds optional background image support in nav and footer section.


daattali:
I'm guessing you're using patterns rather than actual images?

OCram85:
> I'm guessing you're using patterns rather than actual images?

You are absolutely right. Sorry for the misunderstanding.

daattali:
I never tried using frontmatter in css (as part of my philosophy of keeping things simple, but perhaps it's time to start using its benefits) - does it work in github pages? If you look at the css file that gets served to the site, it shows the post-processing file? 

And my usual question: do you have an example of this on your site so I could test it?

OCram85:
Here is a live example on [my blog](https://ocram85.github.io). Additionally I tested the github-pages build with a repo fork.

daattali:
Thanks, I'll take a look next week. I'm traveling and without a laptop so
it's hard to do these things at the moment

daattali:
I just tested this, and while it does seem to work, using frontmatter in the css seems to cause some weird issues. Look at the resulting css file: https://ocram85.github.io/css/main.css

It essentially creates an HTML page. And some CSS rules aren't being respected, for example the `body { ... }` ruleset isn't being applied at all. If there's a way to make the CSS file remain a regular CSS file but still have jekyll process it for variables, then I'd love to add this, but in its current form I wouldn't accept this.

BTW I added your name as a contributor to the readme

OCram85:
> I just tested this, and while it does seem to work, using frontmatter in the css seems to cause some weird issues. Look at the resulting css file: https://ocram85.github.io/css/main.css

You are absolutely right. I didn't notice the html header in the processed css file. Maybe I'll find an similar easy and working solution ^^

OCram85:
According [this](https://github.com/jekyll/jekyll/issues/2730) the solution should be:

## main.css YAML
```yaml
layout: null
```

daattali:
Awesome! Have you tested this out and made sure it really does result in a pure CSS file?

OCram85:
I hope so 🙏 
I've already pushed it to my blog https://ocram85.github.io/css/main.css

main.css

daattali:
Yup it does seem to work, I'm glad I learned about this, thanks!

PR #257, opened 2017-10-03
Fixes mixed EOL Sequences

OCram85:
This is just a cosmetic change:
- Changed EOL sequence from CRLF to LF.
- Trim tailing whitespace in changes files.


daattali:
This isn't too bad at all, I was expecting thousands of lines to have to be updated! Thanks!

OCram85:
It affected about 30 files, but the eol where tracked only these. It depends on the git core settings `core.autocrlf=false` abnd `core.eol=lf`. You can configure git to convert the eol for checkin and checkout actions.

Issue #254, opened 2017-09-22
Make all parameters that use file paths/URLs have consistent and correct logic


daattali:
There are several places in the codebase that accept a path/URL. For example: the links in  the navbar (defined in the config file), the "bigimg" and "image" YAML parameters, the "avatar" config parameter, the "title-img" config param, the "navbar-img", "footer-img" and "page-img" config params (which are new thanks to you!), and maybe a few more that I can't remember right now.

The problem is that right now these parameters all work differently. Some of them only support local relative paths (such as "/img/foo.png") and some pinky support external URL. For local path, extra care needs to be taken to ensure that site.baseurl is always prepended, because otherwise it wouldn't work for GitHub Project Pages. For URLs, both http and https need to work. The different parameters support different combinations of these features, but they should all be fixed to work fully. I think that part of the problem is that each parameter uses its own logic for this, so it'd be a good idea to make a tiny "include-uri" template (as in, actually create a file under _includes/), make that template suppirt all 4 cases (local path that has an empty site.baseurl, local path that has a site.baseurl, http url, https url), and use that template for all the parameters. 

Code samples:
- the navbar menu items do take care of all 4 possibilities I think https://github.com/daattali/beautiful-jekyll/blob/master/_includes/navbarlink.html 
- bigimg param works for local images but not for URLs https://github.com/daattali/beautiful-jekyll/blob/master/_includes/header.html#L11
- image param doesnt work for URLs when site.baseurl is not empty

All these instances need to be consolidated into one. This problem became apparent in https://github.com/daattali/beautiful-jekyll/issues/249

sumitgpt:
Not an expert. But what if we just check if the image link like `bigimg` starts with "/". With Jekyll, you add images in assets and this folder is at the root level. So every local image will start with "/" so that image is looked up from the root. If the link starts with anything else we can consider it an external link.

Now coming to `https` & `http`. I've added `https://` in attribute `url` in my config file. So all the absolute URL of my site are built with `https`. I'd want users to add the preferred version of their site's url in `url` attribute itself. This prevents unnecessary redirects. For an external URL, we can expect that user would add the URL with `http` or `https` in attributes like `bigimg`.

daattali:
I think what you describe in the first paragraph is a similar idea to what I have in the code in some places, it just needs to be abstracted/generalized better so that it can be used everywhere that uses URLs/paths. I would prefer against requiring users to set a `url` config, all these things should work with minimal user intervention

[...]

PR #258, opened 2017-10-03
Changing og:image to absolute URL

sumitgpt:
Relative URL can cause problems with some website. Absolute URLs are recommended.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -84,7 +84,7 @@
   {% endif %}
 
   {% if page.share-img %}
-  <meta property="og:image" content="{{ page.share-img }}" />
+  <meta property="og:image" content="{{ site.url }}{{ page.share-img }}" />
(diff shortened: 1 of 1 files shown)


daattali:
You're absolutely right. In fact I have an issue #254 open for this, because this needs to be fixed in multiple places. But it may not be as simple as simply adding `{{ site.url }}` as you have done; this will work for local images that are hosted in the repo, but it will break images that are external and use an http/https URL. Does that make sense?

daattali:
@OCram85 this is that big issue that I told you about that I think is important :)

OCram85:
@daattali I try working on a solution as soon as possible.

daattali:
No rush!

daattali:
@sumitgpt closing this PR since it does not solve the larger issue. Feel free to submit a PR for #254 :)

PR #265, opened 2017-10-05
Badge functionality to layout/post.html

yonicd:
add badges functionality to post.html and instructions in readme yaml section

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,21 +87,6 @@ subtitle: Here you'll find all the ways to get in touch with me
 ---
 ```
 
-You can also add `Github` and `CRAN` status badges.
-
-  - Github
-    - Set the user name and repository (`gh-repo` field)
-    - Select which badges to display and the order (`gh-badge` field)
-
-```
----
-layout: post
-title: Post Title
-gh-repo: <github username>/<repo name>
-gh-badge: [follow, star, issue, watch, fork, download]
----
-```
-
 You can look at the top of [`aboutme.md`](./aboutme.md) or [`index.html`](./index.html) as more examples.
 
 **Important takeaway: ALWAYS add the YAML front matter, which is two lines with three dashes, to EVERY page. If you have any parameters, they go between the two lines.**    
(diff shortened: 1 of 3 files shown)


daattali:
I just tested this, but it looks like some CSS (or perhaps Javascript that loads CSS?) is needed to make the github buttons work. Did you ever add any css/js for this, other than the HTML? Currently it's just literally printing the words "Star" "Issue" "Follow" "Watch" etc

yonicd:
it should have to need any additional css/js since it works off the same idea of adding a badge to any GH md page.

daattali:
I implemented this on my site successfully. I used the other library, the one that uses iframes directly, because it's way faster.  ghbtns.com

I also changed the `gh_repo` and `gh_badge` to `gh-repo` and `gh-badge` to be more consistent with other YAML params. And I added a bit of CSS so that on small screens the buttons will be one per line because it looks awkward otherwise. Here are the changes:

https://github.com/daattali/daattali.github.io/blob/master/_layouts/post.html#L11-L32
https://github.com/daattali/daattali.github.io/blob/1c9ff4d6ce9d49af84eaf4d4a9cbfb3533319dac/css/main.css#L441-L449


Let me know if you want to update your PR or if I should just do it (I don't mind doing it myself, but it means your name will not show up in the git history)

daattali:
oh right, I actually prefer to exclude those, because I don't want the theme to have R-specific features

yonicd:
no problem. i'll keep it on mine (it's an option anyways that doesnt always have to be used)

yonicd:
the second link is to a different branch so the main.css is different than to the one the PR is connected to.

yonicd:
updated PR to iframes

daattali:
yes the css link is on my own website which has diverged significantly. but it shouldn't matter, it's just 2 css rules that need to copied over

daattali:
Nevermind, you did see it and incorporate it. Thanks.

Can you remove the code for CRAN (as opposed to commenting it out), and also the documentation for this should be within the YAML frontmatter parameters table. There's a big table in the readme with all the available params. Then ready to merge :)

PR #272, opened 2017-10-20
adds editor config for vscode

OCram85:
The workspace related config for the VSCode editor helps:
- VSCode Users
  - Enforce `\n` as end of line sequence for new files and windows users
  - auto trim whitespaces
  - adds final new line
  - shows ruler at char pos 116 (best practice)
  - tabs auto convert into 4 spaces
  - enables git bash as default terminal
  - set default encoding to `utf8` without BOM


daattali:
I actually prefer not to do this PR. These settings are indeed good for
anyone who uses that editor, but every person uses a different editor and I
don't want to gather files that are specific to how someone prefers to
develop. I'm closing this PR, but thank you anyway!

Issue #293, opened 2017-12-26
Invisible avatar is still clickable in .top-nav-short


alefi87:
When scrolling so the avatar fades out and the main nav turns "short" the avatar is still clickable. You can't see it, but the hitbox is still there, so it's a bit confusing, not to mention that it might hijack clicks on some of the content under it (e.g. another link in the text).

I guess that one is tough to solve with pure CSS since `display` is not affected by `transition` rules, but since JS is used for certain functionality, perhaps the avatar animation might as well be done with something like `jQuery('.avatar-container img').fadeOut(500)`?

daattali:
You're right that's a behaviour I've known about for a while but never
seemed to actually bother anyone. If you find a clean way to fix it, you're
welcome to submit a PR!

alefi87:
Submitted a PR. Disregard the git weirdness above - it's just one commit and minor changes to two files.

daattali:
thank you!

PR #295, opened 2017-12-31
Avatar now fades with jQuery (#293)

alefi87:
With some automatic whitespace cleanup courtesy of Atom.


daattali:
I initially tried a CSS-only approach to try to be nice to browsers that disable JS. But now, 3 years later...... nah, they don't matter :)

PR #308, opened 2018-01-19
Add instructions for local builds with bundler

martinjhnhadley:
The current instructions required Docker to be installed and setup which adds complexity to the process. GitHub Page's guide to using bundler to locally build Jekyll sites work with minimal fuss :)

I completely understand if you want to refuse this as adding additional details on how to do local builds invites additional questions when you really don't want to answer them!

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,9 +203,60 @@ Want your website featured here? [Contact me](http://deanattali.com/aboutme#cont
 | [ocram85.github.io](https://ocram85.github.io) | Marco Blessing | A personal blog about PowerShell and automation |
 
 
+## Advanced: Local development
+
+Beautiful Jekyll is meant to be so simple to use that you can do it all within the browser. However, if you'd like to develop locally on your own machine, that's possible too if you're comfortable with command line. There are two different options available which are briefly detailed below.
+
+Disclaimer: I personally am NOT using local development so I don't know much about running Jekyll locally. If you follow either route, please don't ask me questions because unfortunately I honestly won't be able to help!
+
+### Bundler
+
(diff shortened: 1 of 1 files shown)


jennydaman:
Local development involves use of CLI, no matter what you do. After making the assumption that the user has CLI experience, I wouldn't say that experience with Docker is less common than experience with ruby.

Anyway, I personally disagree with running bundler directly on my own machine.

Gem dependency and version management is poor, after running bundle install you end up with much environment pollution. The alternative is to install gems directly to $CWD, though this takes some tinkering.

The whole point docker is to achieve two objectives: isolation and cross-compatibility.

I think the philosophy is: markdown is so easy, you don't need to build and run locally. But if you have the technical background, you can figure anything out yourself.

daattali:
While I don't 100% agree with @jennydaman because introducing Docker *can* be more complex for users who don't already use Docker but do have CLI familiarity, what he said about the philosophy is correct :) I would prefer not to add a few more paragraphs catering to advanced users, because if they can follow those instructions, then chances are they can fairly easily figure them out on their own too. I would be ok with adding a link to the github help page (can be a new PR), but wouldn't want this entire section.

jennydaman:
Random idea: perhaps documentation should be migrated to Github Wiki, instead of one huge README.md file. (And there can be a page dedicated to advanced topics, too)
-------- Original Message --------

daattali:
@jennydaman I think github wiki pages make sense for a large complex project where lots of documentation about API, usage, examples, etc is required. With this theme, I purposely want to keep it simple and it is only a theme so I wouldn't want the documentation to become too large to require multiple pages. I really strongly believe in simplicity and not overloading with functions and info :)

PR #310, opened 2018-01-19
Adds Matomo (aka Piwik) integration

### LINKED ISSUE #309: Piwik / Matomo integration
I use my personal blog with Piwik / Matamo stats.
Is there anyone interested in a PR for including this to beautiful-jekyll like GoogleAnalytics?

OCram85:
## :question: About
Some users don't like using Google Analytics. Alternatively they use the open-source tracking suite Matomo (aka Piwik). I added 3 simple config keys to set up the tracking with an existing Matomo instance.

All you need is:

- Matomo Site ID
- Matomo Instance URI
- Enable / Disable the Opt-Out feature.

## :notebook: Changes
- Restructured the config keys section in `_config.yml`. There is now a dedicated *Web Statistics Section*
  - contains Google and Matomo preferences
- added `matomo.html` include file with the JS tracking code
- added Do-Not-Track link in footer if Opt-Out is set. This will open a new page in the browser language.
- tracking code in header headed if keys set.

## :bulb: Improvements
@daattali : Maybe you have got an other idea where and how to add the Do-Not-Track link.

## :link: Ref
fixes #309

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -32,6 +32,13 @@
       &nbsp;&bull;&nbsp;
       <a href="{{ site.url }}">{{ site.url-pretty }}</a>
       {% endif %}
+
+      {% if site.matomo %}
+      &nbsp;&bull;&nbsp;
+        {% if site.matomo.opt-out %}
+          <a href="http://{{- site.matomo.uri -}}/index.php?module=CoreAdminHome&action=optOut" target="_blank" class="text_muted">Do-not-Track</a>
+        {% endif %}
+      {% endif%}
       </p>
           <!-- Please don't remove this, keep my open source work credited :) -->
     <p class="theme-by text-muted">
--- a/_config.yml
+++ b/_config.yml
@@ -101,15 +101,29 @@ share-links-active:
 # Remove this if you don't want a link in the footer
 url-pretty: "MyWebsite.com"  # eg. "deanattali.com/beautiful-jekyll"
 
-# --- Misc --- #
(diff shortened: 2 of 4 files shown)


daattali:
Can I see this in action on one of your websites? Especially the opt out behaviour?

OCram85:
I simplified the integration. So the config is different than on my instance. But you can follow this link and use the opt out link in the English section. https://ocram85.github.io/impressum

_includes/footer.html
```diff
@@ -32,6 +32,13 @@
       &nbsp;&bull;&nbsp;
       <a href="{{ site.url }}">{{ site.url-pretty }}</a>
       {% endif %}
+
+      {% if site.matomo %}
+      &nbsp;&bull;&nbsp;
```
daattali: This whitespace-bullet-whitespace should go inside the next if statement, right? Otherwise, if you have matomo but don't offer an opt out, there will be nothing after the last bullet?

daattali:
So this will add a "Do-Not-Track" link in the footer, similar to your "Impressum" link, but only if the user chooses to allow opt out. Is that correct?

OCram85:
Yes, the Piwik tracking will be enabled for all by default. But each user can disable it for his browser session by using the the "Do-Not-Track" link. This sets a cookie which tells piwik not to track.

OCram85:
... And yes I decided to put the "Do-Not-Track" link in the footer part. I hope you've got a better idea where to place it^^

daattali:
I can't think of anywhere better. That's probably the least intrusive

PR #315, opened 2018-01-28
changed css to scss to allow easier ports. liquid now just sits at the top.

shyd:
Hi,
I love your theme! I am porting this theme for local development with gulp task runner. Therefor I don't want to use liquid inside css. So I decided to switch over to scss and put the color and image settings into scss without breaking your architecture.

I just want to drop my changes here. Maybe this is useful if someone has something like me in mind and wants to get rid of liquid in css as well. This is a good starting point, because only the definitions have to be changed at the top of the file.

Cheers 🍻

### DIFF
--- a/css/main.scss
+++ b/css/main.scss
@@ -4,6 +4,39 @@ layout: null
 
 @import url("pygment_highlights.css");
 
+$page-col:{{ site.page-col }};
+$link-col:{{ site.link-col }};
+$hover-col:{{ site.hover-col }};
+$navbar-col:{{ site.navbar-col }};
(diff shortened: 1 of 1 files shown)


daattali:
Thanks for the initiative @shyd ! I've thought about making the switch to scss in the past, but I decided against it in favour of keeping it a bit more familiar to most people. I realize the syntax itself is not any more difficult, but I think simply seeing a `.scss` file ending will scare off people who are not experienced, so I'd rather keep it as-is. I do appreciate the PR :)

PR #322, opened 2018-02-17
Container now have all proper dependencies

JosemyDuarte:
Container was not working because didn't have all dependencies, and change CMD command for ENTRYPOINT.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,12 +2,11 @@ FROM jekyll/jekyll
 
 WORKDIR /srv/jekyll
 COPY Gemfile .
+COPY Gemfile.lock .
 
 RUN gem install bundler
-RUN bundle install --clean --quiet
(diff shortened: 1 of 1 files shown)


JosemyDuarte:
The former Dockerfile didn't work for me, container displayed "Could not find i18n-0.9.1 in any of the sources". With this changes all worked fine, hope this helps.

daattali:
Thanks for the PR @JosemyDuarte 

As I mention in the README, I haven't tested the docker setup, and I rely on trusted contributors for that

@jennydaman how does this look? I'm pretty sure the full long "bundle exec jekyll server" with port and host params is wanted. I'm not sure if including the lock file is standard practice or not, do you have any comments?

daattali:
Thanks for your input @jennydaman and thanks for the PR @JosemyDuarte !

PR #323, opened 2018-02-18
Generate sitemap.xml

JosemyDuarte:
Automatically generate sitemap.xml to use with Google Search Console and improve SEO.


daattali:
Thanks!

Issue #324, opened 2018-02-20
upgrade front awesome version 5 


allan9595:
Hello, i try to add discord font awesome on my site but seems like Beautiful jekyll still use font awesome version 4 so my discord icon not works. Could you update it to version 5 for font awesome?

daattali:
Sure I can look into that when I have a few minutes. I would also accept a
pull request

allan9595:
Thank you !

daattali:
Looks like this will have to wait a bit. fontawesome just released version 5.0.6 17 days ago, and it looks like the CDNs serving fontawesome haven't updated yet (which I totally understand if it's only been 17 days). I checked https://www.bootstrapcdn.com/fontawesome/ and https://cdnjs.com/libraries/font-awesome both of them are still serving 4.x

I'll check again periodically and update when it's available. In the meantime feel free to change it locally on your project to the new version by actually downloading the fontawesome files and referring to the local copy rather than to an online copy

shyd:
I have done this in my fork in order to use the docker icon 🐳  But I use gulp task runner with bower to build my site. <https://schuett.io>
I will create a PR with the icon changes later. Then you can merge it if v5 becomes available at CDNs.

Cheers

OCram85:
@shyd: Awesome Blog! Super Arbeit :+1:

allan9595:
@shyd Do you still able to use the rss feed after you updated to the font awesome 5? My site runs into a rss feed issue now. (https://untcybersecurity.com/feed.xml)

shyd:
@allan9595 Yes, it's stable for me. I just branched my feature branch into gh-pages for demonstration: <https://shyd.github.io/beautiful-jekyll/feed.xml>

andreas-paul:
Is there any update on the migration to Font Awesome v5? I tried to implement it myself but I can't get the icons of v5 to work. Seems like the cdn's are still on v4 which is quite strange.

daattali:
Not going to be implemented until I have a good reason to

daattali:
You can look at [this PR](https://github.com/daattali/beautiful-jekyll/pull/400) by @shyd if you want to upgrade!

daattali:
Done

PR #325, opened 2018-02-23
Updated to Font Awesome v5, added docker social icon

shyd:
Hi,
as promised in #324, here is my PR with all needed updates in order to use font awesome v5. I also added docker to the SocialNetworks.yml
Some icons changed a bit since older versions in size and the icon itself.

Cheers 🍻

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -76,6 +76,7 @@ social-network-links:
   facebook: daattali
   github: daattali
   twitter: daattali
+#  docker: yourname
 #  reddit: yourname
 #  google-plus: +DeanAttali
 #  linkedin: daattali
--- a/_data/SocialNetworks.yml
+++ b/_data/SocialNetworks.yml
@@ -9,89 +9,94 @@
 facebook:
   name: "Facebook"
   baseURL: "https://www.facebook.com/"
-  icon: "fa-facebook"
+  icon: "fab fa-facebook-f"
 
 github:
   name: "GitHub"
   baseURL: "https://github.com/"
-  icon: "fa-github"
+  icon: "fab fa-github"
 
 twitter:
   name: "Twitter"
   baseURL: "https://twitter.com/"
-  icon: "fa-twitter"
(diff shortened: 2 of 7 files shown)


daattali:
From the 2-minute research that I did, it looked like you don't **have to** switch to SVG. I prefer to stick with the old version for now that doesn't require a javascript file and handles it at in css. Are there big advantaged to moving to svg?

daattali:
From the 2-minute research that I did, it looked like you don't **have to** switch to SVG. I prefer to stick with the old version for now that doesn't require a javascript file and handles it at in css. Are there big advantaged to moving to svg?

shyd:
Agree. But:
Using js is the recommended method to use v5. The big advantage I see, is that there is only one request needed to get the fonts: fetching the js-file. When using css there are more requests needed, because the browser needs to fetch the webfonts in addition to the css-file.
Nevertheless using css doesn't require as much traffic as using js. Where js includes every icon, css downloads needed fonts only.

My PR works either way, just include css or js:

```
common-ext-css:
  - "//use.fontawesome.com/releases/v5.0.6/css/all.css"

common-ext-js:
  - "//use.fontawesome.com/releases/v5.0.6/js/all.js"
````

abelcheung:
I'm personally switching to JS version for testing:

Pro: It offers some cool effect not available with webfont + CSS, like arbitrary resizing, rotating, and icon masking (allowing background image to show through icon layers, for example)

Con: It can slow down IE to a crawl with certain usage ([specifically with CSS pseudo elements][1]),
which has no fix for now, except using webfont + CSS version

The choice to make really depends on target audience; but for generic usage I'm siding with webfont version too.

[1]: https://stackoverflow.com/questions/50976897/fontawesome-5-svg-icons-data-search-pseudo-elements-causes-100x-slowdown-in-ren

daattali:
Thanks for your input @abelcheung . For now I'll stick with this as you said, may revisit in the future

PR #335, opened 2018-03-15
Rename _config.yml to config.yml

hoverzheng:


daattali:
Why this rename?

daattali:
It needs to remain _config https://jekyllrb.com/docs/configuration/

PR #341, opened 2018-04-06
adding scroll to top in structure

atmosmps:
I've added a simple scroll button to the top, inside the framework to improve user navigation on articles and other pages.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -165,6 +165,9 @@ defaults:
 # Use tags index page and make tags on each post clickable
 link-tags: true
 
+# Enable scroll button to the top of the page
+scroll-to-the-top: false
+
 # Exclude these files from production site
 exclude:
   - CHANGELOG.md
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -1,7 +1,4 @@
 <footer>
-  <!-- Return to Top -->
-  <a href="#" id="return-to-top"><i class="fa fa-arrow-up"></i></a>
-
   <div class="container beautiful-jekyll-footer">
     <div class="row">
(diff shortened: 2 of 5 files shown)


daattali:
Thanks for the PR, I appreciate it. Before adding any feature to beautiful jekyll I make sure it's extremely robust and works on different environments. I have very strict standards for accepting new features because this theme is used by so many people unfortunately.

I tested this feature on your website briefly and noted it has some UI/UX issues: When I'm on mobile, after clicking on it once the arrow gets stuck at the up position. On my desktop if I make the window narrow and scroll all the way down, the up button is not clickable anymore. This should be a customizable feature that should have a setting in the config file. The arrow isn't centered vertically. After clicking on the arrow, while the page is scrolling up, the background of the circle flickers.

If you can address all these issues and make sure the UX with this button is excellent then I'll be happy to accept the PR!

atmosmaciel:
Hi! Thanks for the feedback, the animations were purposeful, but really recognize that they are not usable from UX's point of view. I apologize for not having done all the necessary tests and will make the correct corrections of this feature as soon as possible. Thanks again for your patience and observations.

atmosmaciel:
The code writing got bigger, but I think the adjustments were better applied now. When the user is rolling the page the arrow will fade to lose focus and only when triggered will focus again. The configuration option in ```_config.yml``` has also been added to make this setting optional. The functionality now is also responsive.

daattali:
@atmosmps I really appreciate your PR. I try to be very mindful of feature creep in order not to bloat the codebase too much, and this seems like it's adding a lot of code to maintain for a relatively small feature. If there was an elegant solution that was very lightweight I would consider it, but I prefer not to add this much code. I do appreciate your energy here

PR #351, opened 2018-04-25
Readme.md: fix instructions for dirs with spaces

ck37:
$PWD needs to be surrounded in quotes so that the commands work when any folder in the path has a space in it.

Currently if someone runs the command as-is, and their folder path has a space in it, they will get the build error message `"docker build" requires exactly 1 argument` build and run error message `docker: invalid reference format` during run.


daattali:
Thanks, good catch!

PR #355, opened 2018-05-11
Notch display

kdo9921:
Notch display(like iPhoneX) support


daattali:
Thanks for the pull request, I missed the notification and haven't seen it until now. I don't know much about new ios, can you explain what these changes do? Will all other browsers remain unaffected?

daattali:
Thank you, as a non iphone user I never would have known

PR #357, opened 2018-05-24
update jQuery

OCram85:
Updates jQuery as first preperation step for GDPR compliance.


abelcheung:
Well, IMHO many of the resources (JQuery, Bootstrap etc) could have been offloaded to CDN instead of being bundled and served from Github. Usage of FontAwesome is already done in this way. That could acutally improve performance as well, since multiple resources are loaded from different locations in parallel.

abelcheung:
Please forget previous comment since I haven't checked out the umbrella task at that time. Looks like more than technical issue have to be considered as well.

abelcheung:
Confirmed, this patch can't be applied. Bootstrap 3.x explicitly don't support JQuery 3, only accepts &ge; 1.9.1 and 2.x.

daattali:
Can't be applied *until bootstrap is updated

danielmpetrov:
> Confirmed, this patch can't be applied. Bootstrap 3.x explicitly don't support JQuery 3, only accepts ≥ 1.9.1 and 2.x.

[Boostrap 3.3.7](https://blog.getbootstrap.com/2016/07/25/bootstrap-3-3-7-released/) introduced support for JQuery 3, so updating to the latest Boostrap 3 version (currently v3.4.1) will do the trick.

daattali:
Closing this for now, if bootstrap will be upgraded then we can revisit

PR #359, opened 2018-05-24
Implement FontAwesome 470

OCram85:


daattali:
Font awesome was upgradd to 5.12

PR #366, opened 2018-06-13
Fix code fence css bug when published on gh pages

### LINKED ISSUE #365: Code fences render differently on the bundled version of jekyll than on Github Pages.
Code fences render differently on the bundled version of jekyll than on Github Pages. 

Github Pages:
image

Locally with Jekyll 3.5.2:
image

I've narrowed the issue to the fact that GHP seems to insert a `<div class="highlight">` before the `<pre class="highlight">` which triggers a rule in `main.css` that sets `padding: 0`. 

```
.highlight > pre {
  padding: 0;
}
```

hkhanna:
fixes #365

### DIFF
--- a/css/main.css
+++ b/css/main.css
@@ -688,6 +688,9 @@ td.gutter pre {
 .highlight > pre {
   padding: 0;
 }
+.highlight > pre.highlight {
+  padding: 9.5px;
+}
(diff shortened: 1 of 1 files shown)


daattali:
thanks!

PR #374, opened 2018-07-06
Add gtag support for Google Analytics

hkhanna:
Google Analytics has a new type of default `<script>` tag called `gtag`. This PR adds support for that.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -103,6 +103,9 @@ url-pretty: "MyWebsite.com"  # eg. "deanattali.com/beautiful-jekyll"
 
 # --- Web Statistics Section --- #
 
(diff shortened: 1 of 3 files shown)


JosemyDuarte:
Does this make useless the former way to put Google Analitycs ID? I meand about this lines:

```
# Fill in your Google Analytics ID to track your website using GA
# google_analytics: ""
```

Or they are different things?

hkhanna:
@JosemyDuarte they are different things. I think the google_analytics tag is legacy, but many many sites still use it. So I'd keep it in the beautiful-jekyll codebase.

JosemyDuarte:
Great! Thanks for your response.

PR #404, opened 2018-09-21
Update _config.yml

soswift:


daattali:
You need to do these changes on your own copy of the project

PR #414, opened 2018-10-11
Use HTTPS as default protocol of site

mashed-potatoes:
I think that is better to use HTTPS URL..


daattali:
Now that github pages supports https for all its websites, this is a good change, thanks

PR #423, opened 2018-11-10
fix security vulnerabilities

npes87184:
GitHub notify some security vulnerabilities related to jekyll and ffi.
Upgrade the version of jekyll and ffi to the requirement.


daattali:
thanks

PR #437, opened 2018-12-16
Update Jekyll requirements: old {%- format support was removed + site…

gpotter2:
…map -> plugins_dir

Hello & Thanks a lot for the project, we're using it on [scapy](https://scapy.net/)

- Recent jekyll versions have stopped supporting the `{%-` format, and now require the `{%` format. While most of the code already has the new format, this PR cleans up the last bits.
- `jekyll-sitemap` is now required to go in `plugins_dir` rather than `plugins` (according to warnings when building)

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -237,9 +237,11 @@ prose:
           label: "Publish"
           help: "Check to publish post, uncheck to hide."
 
+plugins_dir:
+  - jekyll-sitemap
+
 plugins:
   - jekyll-paginate
-  - jekyll-sitemap
 
 # Beautiful Jekyll / Dean Attali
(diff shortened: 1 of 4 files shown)


daattali:
Thanks @gpotter2 

Do you have any source that says `{%-` is becoming unsupported? I couldn't see anything to support that claim. Same with `plugins_dir` - I don't see anything in the sitemap' repository or other places I looked that suggest this

gpotter2:
To be fair, I only based this PR on what Jekyll gave me as an output, but that was on a computer running WSL which I don’t have access to.

I think the PR is wrong about the dashes, but I could

PR #440, opened 2018-12-20
[Feature] Staticman v3 Support

### LINKED ISSUE #115: Can anyone add staticman comment system?
Hello, I've found a wonderful static comment system Staticman ([https://staticman.net/](https://staticman.net/)). Can anyone add it to our Beautiful Jekyll? :D Thank you very much!

VincentTam:
Resolves #115.  A port of [Minimal Mistake](https://github.com/mmistakes/minimal-mistakes)'s Staticman support.  CSS has been compiled from SCSS with an online compiler, so there may be some unused properties.

Due to Staticman's public API instance bottleneck eduardoboucas/staticman#227, the default API endpoint is set to the one associated with @staticmanlab, my public API instance for GitLab.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -129,6 +129,15 @@ url-pretty: "MyWebsite.com"  # eg. "deanattali.com/beautiful-jekyll"
 # To use Facebook Comments, fill in a Facebook App ID
 # fb_comment_id: ""
 
+# Staticman support
+staticman:
+  repository     : # GitHub username/repo-name e.g. "daattali/beautiful-jekyll"
+  branch         : # "master"
+  endpoint       : # URL of your own deployment with trailing slash, will fallback to the public GitLab instance
+reCaptcha:
+  siteKey        :
+  secret         :
+
 # --- Misc --- #
 
 # Facebook App ID
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,7 @@
 The MIT License (MIT)
 
 Copyright (c) 2016 Dean Attali
+Copyright (c) 2017 Michael Rose
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/README.md
+++ b/README.md
@@ -104,13 +104,20 @@ Many personalization settings in `_config.yml`, such as setting your name and si
 
 ### Allowing users to leave comments
 
-If you want to enable comments on your site, Beautiful Jekyll supports either the [Disqus](https://disqus.com/) comments plugin or [Facebook](https://developers.facebook.com/docs/plugins/comments) comments.
+If you want to enable comments on your site, Beautiful Jekyll supports either the [Disqus](https://disqus.com/) comments plugin, [Facebook](https://developers.facebook.com/docs/plugins/comments) comments or [Staticman](https://staticman.net).
 
 To use Disqus, simply sign up to [Disqus](https://disqus.com/) and add your Disqus shortname to the `disqus` parameter in the `_config.yml` file.
 
 To use Facebook comments, create a Facebook app using [Facebook developers](https://developers.facebook.com/docs/apps/register), and add the Facebook App ID to the `fb_comment_id` parameter in `_config.yml`.
 
-If either `disqus` or `fb_comment_id` parameters are set in the configuration file, then all blog posts will have comments turned on by default. To turn off comments on a particular blog post, add `comments: false` to the YAML front matter. If you want to add comments on the bottom of a non-blog page, add `comments: true` to the YAML front matter.
+To use Staticman, send an invitation to the public GitLab instance @staticmanlab, click
+
+    https://staticman3.herokuapp.com/v3/connect/github/username/repo-name
+
+to programmatically accept your invitation.  Then fill in your `repository` and
+`branch` in `_config.yml`.  The `branch` in `staticman.yml` should match the one in `_config.yml`.
+
+If either `disqus`, `fb_comment_id` or `staticman` parameters are set in the configuration file, then all blog posts will have comments turned on by default. To turn off comments on a particular blog post, add `comments: false` to the YAML front matter. If you want to add comments on the bottom of a non-blog page, add `comments: true` to the YAML front matter.
 
 ### Adding Google Analytics to track page views
 
--- a/_data/ui-text.yml
+++ b/_data/ui-text.yml
@@ -1,0 +1,494 @@
+# User interface text and labels
+
+# English (default)
+# -----------------
+en: &DEFAULT_EN
+  comments_label             : "Leave a comment"
+  comments_title             : "Comments"
+  comment_form_info          : "Your email address will not be published. Required fields are marked"
+  comment_form_comment_label : "Comment"
+  comment_form_md_info       : "Markdown is supported."
+  comment_form_name_label    : "Name"
+  comment_form_email_label   : "Email address"
+  comment_form_website_label : "Website (optional)"
+  comment_btn_submit         : "Submit comment"
+  comment_btn_submitted      : "Submitted"
+  comment_success_msg        : "Thanks for your comment! It will show on the site once it has been approved."
+  comment_error_msg          : "Sorry, there was an error with your submission. Please make sure all required fields have been completed and try again."
+en-US:
+  <<: *DEFAULT_EN
+en-CA:
+  <<: *DEFAULT_EN
+en-GB:
+  <<: *DEFAULT_EN
+en-AU:
+  <<: *DEFAULT_EN
+
+# Spanish
+# -------
+es: &DEFAULT_ES
+  comments_label             : "Dejar un commentario"
+  comments_title             : "Comentar"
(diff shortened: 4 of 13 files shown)


daattali:
Thanks for adding the code and instructions. In the instructions, "send an invitation to the public GitLab instance @staticmanlab" - can you explain what that means, what needs to be done? Does clicking on the URL below it perform that invitation, or is that the next step?

VincentTam:
Excuse me for the lack of clarity.  A picture like

staticmanlab invitation on GitHub

would illustrate the idea of "invitation" much clearer.  The next step is to fire an HTTP GET request to the URL.

The above setup are, in fact, parallel with the with setting up the site & repo config YML files.  However, to organize them in reading order, I prefer putting the GitHub invitation first because the response from the Staticman API instance gives users a sense that the static comment service is working.

This instructions echoes those on Staticman's official website, except that I've replaced them with my public instance because of the linked Staticman issue. (It's reported that the project owner's public API instance can no longer accept invitation.)

N.B. My current Staticman API instance is hosted on a free Heroku dyno, which should *never* be used for production.  Consequently, the server idles after 30 minutes of inactivity.  When a comment is sent while the server is idling, it is, luckily, not lost, but subject to an expected delay of 15-20s for server restart.  One always has the possibility to [host his/her owner API instance](//flyinggrizzly.io/2017/12/setting-up-staticman-server/).  The original intention to build my own one is to test Staticman's native GitLab support, so that fits role of a free Heroku dyno.  It's a *coincidence* that custom API servers provide a *workaround* for the linked Staticman issue.

daattali:
Thanks for all your work. I've finally given staticman a quick look and it does look very useful, and I see that you built that bot for everybody's help, so thank you from everyone else too :)

I haven't looked into the PR in too much detail yet, I have a few basic questions first:

- Why is it called v3 if staticman is at v2? Do you mean v3 because it's using your API rather than the official one?
- To make sure i understand: first I need to add staticmanlab as a collaborator, and then the next step is to go to the heroku URL?
- In the config.yml file, what is recaptcha for, is it mandatory when using staticman, and how do you fill out its values?
- If the config.yml file stays as is, with the staticman parameters existing but set to empty strings, would all the staticman css and all associated html/JS be included on the rendered pages?

VincentTam:
Thanks for your thoughtful questions.

The [source code](https://github.com/vincenttam/staticman) (`deploy`  branch) of my API instance is forked from ntsim/staticman `dev` branch, so my instance includes eduardoboucas/staticman#219 (and 231 as well).  The project owner did build a public testing instance https://test.staticman.net, whose displays "Hello from Staticman version 3.0.0!".  However, he *didn't* (and perhaps *won't*) disclose its associated GitLab account.  As a result, I've responded to his question by building a public GitLab instance.

I'm using v3 because this shows that v3 works for GitHub, and indicates that this also works for GitLab.  "v3" is the version number found in Nicolas Tsim's code merged against the `dev` branch of the official repo.

Yes for the second question, with a remark that the URL is *not* Heroku specific.  Staticman v3 URL scheme is documented in the linked PR.

The role of reCAPTCHA in config.yml is to provide optional support for this great tool.  While writing README, I've forgotten to add an important point about reCAPTCHA.  The site secret has to be encrypted by

    GET v3/encrypt/SITE_SECRET

"No" for the last question.  You'll see that I'm finally using the same logic as disqus.html: empty parameters will evaluate to false.

daattali:
Thanks for all the details.

Regarding the recaptcha - if it's for use within staticman, do you think it would make sense to add it as a parameter nested inside staticman instead of its own parameter? And in order to not overload the README, maybe add a comment in the config file about howto fill out the recaptcha params?

daattali:
I've modified the README a bit, can you also make sure what I wrote is correct and makes sense? I haven't tried this feature yet, I wrote it based on my understanding.

VincentTam:
I'm neutral on your suggestion about refactoring `site.recaptcha` to `site.staticman.recaptcha`.  Beautiful Hugo, the Hugo port of this theme has actually implemented that, whereas I'm following Minimal Mistakes's config.yml.  I *speculate* that Michael Rose has left `recaptcha` outside `staticman` for the possible use of other comments provider(s) in the future.

If you want `recaptcha` nested inside `staticman`, replacing all instances of `site.recaptcha` with `site.staticman.recaptcha` will do.

I've your edit on README clearer than mine.  I would suggest adding a screenshot like

the one in Staticman's official documentation

since README is to get users set up the blog in minutes, and human minds read pictures much quicker than text.

I've missed an important point about encrypting reCAPTCHA site secrets.  I'll edit config.yml template this evening.  (I'm replying on mobile.)

_config.yml
```diff
@@ -129,14 +129,15 @@ url-pretty: "MyWebsite.com"  # eg. "deanattali.com/beautiful-jekyll"
 # To use Facebook Comments, fill in a Facebook App ID
 # fb_comment_id: ""
 
-# Staticman support
+# Staticman
```
VincentTam: Changed this line by mistake.

VincentTam:
Modified `_config.yml` and `staticman.yml` to give clearer instructions about reCAPTCHA site secret encryption via `v3/encrypt`, and added a screenshot on collaborator invitation.

daattali:
Since the recaptcha is only used in staticman for now, and the instructions for it also point to a staticman resource for encrypting the secret, I do prefer it to be a nested parameter under staticman in the config file. Why do the recaptcha parameters have to be repeated twice, once in _config.yml and once in staticman.yml?

VincentTam:
The client-side `_config.yml` is for Jekyll templates; `staticman.yml` is for Staticman API server.

In Staticman v2, Staticman API server only recognises root level `staticman.yml` in a GitHub repo.  That's documented in [step 2 of the Git repo setup](https://staticman.net/docs#step-2-create-a-configuration-file).

To understand the separate roles that the two config files, you may view Nicolas Tsim's sample repo at https://gitlab.com/nstim/test-staticman (its read access requires a GitLab account) provided in Staticman PR 219.

daattali:
Great, thanks for bearing with me through this.

Last few comments:
- Please add this feature and yourself to CHANGELOG.md
- Can I add in config.yml in the comment for the recaptcha param something along the lines of "If you use recaptcha, you must also set these parameters in staticman.yml"?
- The CSS file has a lot of rules with very generic selectors. This will override many css rules on the entire page. Please prepend all css selectors with `.staticman-comments `

VincentTam:
Questions are more than welcome.  I'm indebted to Nicolas Tsim's detailed guidance while setting up my public Staticman API instance for [GitLab.com](https://gitlab.com).

Some little technical remarks:

1. I apologize for having inadvertently introduced a BOM at the beginning of the CSS file.  This has been _removed_ at f4fbd0343bd5f94290f2368d38ec42fdb3545c0e on the `gh-pages` branch for [my demo site](https://git.io/bjsm18).  To get a cleaner commit history, I squashed it with 3dcbab360c74cf835a4783ba6d4613c29a2c0d47 to obtain dcaeaf3.  This is _invisible_ on GitHub, but this can be viewed with Git CLI.
2. The commit aab7d05d3e0bc4b87db39739f17ecd100e803786 has introduced a leading whitespace in the change log.  While this makes _no_ difference on GitHub, this may break the syntax highlighting in text editors.  Therefore, I've removed it.

daattali:
Thanks for taking such good care, and removing a whitespace from weeks ago. And I didn't notice the BOM - thanks for taking care of that as well.

daattali:
I merged and tested this out myself on https://daattalitest.github.io/2015-02-28-test-markdown/. Worked really well, thank you!

Two follow up questions:

- Why is email a required field, it doesn't seem to be saved anywhere that's useful even for the repository owner. Is there any way to retrieve the email of a user? If not, I think it should be removed?
- After submitting a comment, I get redirected to a result page that just shows me the result JSON `{"success":true,"fields":{"message":"another try","name":"kkk","email":"ed8b33f09416dd8c64479d3f3aa04465","url":"","hidden":"","date":"2018-12-25T23:10:34.191Z"}}` - this is not useful for users, they would think an error happened. Can we change it so that an alert box says the comment was submitted, or even just refresh the page, or anything else. The way it currently works is not ideal

VincentTam:
Thanks for adopting my PR!  I've left a testing comment on your test site, and I also observed the raw JSON response coming up instead of the UI text "Thanks for your comment! It will show on the site once it has been approved.".  That's a bug in my template that I've never observed from my experience of building [Hugo & Jekyll sites on GitLab](https://gitlab.com/users/staticmanlab/contributed).  Gimme 5 min to fix that.

- The email field will be used for gravatar.
- This display of raw JSON reponse is due to the [omission of the `options[redirect]` field](https://staticman.net/docs/index.html#step-3-hook-up-your-forms) in the HTML form.

    Since I _didn't_ observe that for my test site ([VincentTam/beautiful-jekyll/gh-pages](https://git.io/bjsm0)), I did [a direct comparison between these two branches](https://github.com/VincentTam/beautiful-jekyll/compare/gh-pages..daattalitest:master).  The only observable difference is
  + Project changed to user GitHub page: `master` replaced by `gh-pages`
  + `moderation` switched to false

PR #444, opened 2018-12-26
Fix Staticman's raw JSON response in #440

VincentTam:
Resolves https://github.com/daattali/beautiful-jekyll/pull/440#issuecomment-449878522.

My test site: https://cox-zucker.github.io/2015-02-28-test-markdown/ ([Source](https://github.com/cox-zucker/cox-zucker.github.io))

Staticman comment submitted


daattali:
Great, I tested with this fix and I'm happy now. Thanks!

PR #450, opened 2019-01-20
README.md: add solution of write permission when setup docker env

npes87184:
Signed-off-by: Yu-Chen Lin <npes87184@gmail.com>


daattali:
Thanks for the PR! When would you run into this issue?

npes87184:
> Thanks for the PR! When would you run into this issue?

When using docker with linux (without the more common dinghy on OSX), the docker containers don't have write permission to the bind-mounted volume containing the your git checkout of the code.

PR #472, opened 2019-03-28
Fix printing site.social-network-links

cogniti:
Hello.
If there is no site.social-network-links, do not print ul tag.
Thanks.


daattali:
Won't this change make the ul tag appear many times?

cogniti:
The ul tag appears once.

```
<ul ...>
empty
</ul>
```

cogniti:
`{%- if curkey == 'rss' -%}` was not considered. I'm sorry.

PR #497, opened 2019-05-27
Fix showing background img with non-empty baseurl

abelcheung:
For Github project pages, site background image would fail to load because baseurl is not prepended to image path.

Also switch globally to `relative_url` for all baseurl prepends, which is more robust than custom URL mangling, as `relative_url` would sanitize baseurl before prepending.

### DIFF
--- a/404.html
+++ b/404.html
@@ -9,5 +9,5 @@ permalink: /404.html
   <h1>Move along. (404 error)</h1>
   <br/>
 
-  <img src="{{ site.baseurl }}/img/404-southpark.jpg" />
+  <img src="{{ 'img/404-southpark.jpg' | relative_url }}" />
 </div>
--- a/_data/SocialNetworks.yml
+++ b/_data/SocialNetworks.yml
@@ -78,7 +78,7 @@ telephone:
 
 rss:
   name: "RSS"
-  baseURL: "{{ '/feed.xml' | prepend: site.baseurl }}"
+  baseURL: "{{ '/feed.xml' | relative_url }}"
   icon: "fa-rss"
 
 steam:
--- a/_includes/footer-scripts.html
+++ b/_includes/footer-scripts.html
@@ -12,7 +12,7 @@
 
 {% if page.js %}
   {% for js in page.js %}
-	<script src="{{ js | prepend: site.baseurl | replace: '//', '/' }}"></script>
+    <script src="{{ js | relative_url }}"></script>
   {% endfor %}
 {% endif %}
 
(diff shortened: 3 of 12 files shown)


daattali:
@abelcheung should this change not have introduced any breaking changes?

I just tried forking a github pages Project site after merging your changes and noticed the following paths that didn't seem to be correct: 

- The link at the top left of the navigation bar, that should take you back to the homepage, ignores the baseurl (it just goes back to the user page, but ignores the project name)
- Same with the avatar image - it ignores baseurl
- All the posts in the tags page https://daattalitest.github.io/test/tags/ don't have the baseurl in their paths so none of the links there work 
- The image beside a post has the wrong path (The post "First post!" found on the second page has an image beside it that doesn't have baseurl and is therefore broken https://daattalitest.github.io/test/page2/)

abelcheung:
The first was normal, behavior was the same before code change. The link has a final slash missing, and Jekyll don't know that it should be redirected to `<full_url>`/index.html . On the other hand Github serves the page differently, and it can handle the redirection.

But for the others it's odd, I haven't encountered those problems in personal test at all. Let me take a look and provide you more detail soon.

abelcheung:
Ah my bad, I haven't explained clearly. The merged changes are intended to use with existing project page settings, and not Jekyll recommended ones:

```yaml
#url: "http://username.github.io/projectname"
#baseurl: "/projectname"
```

abelcheung:
After using the above settings, all errors you listed should (hopefully) be gone

daattali:
I feel so silly. You're absolutely correct. The last bullet point, the image path, is still incorrect because it's missing the project name in the path, but I assume that was the case beforehand too (the image feature was added by a third party and it's possible they didn't know they need to take care of the project page case)

abelcheung:
The "First Post" only contains a few lines:

```markdown
---
layout: post
title: First post!
image: /img/hello_world.jpeg
---

This is my first post, how exciting!
```

Do you mean the `image` property in front matter? If that's what you mean, then browser cache need to be cleared :grin:. In your test site, the avatar icon is successfully replaced for that particular post.

abelcheung:
@daattali  Finally understand what's happening now:

image

It sounds odd to me, is that intended behavior to use image both as article image and as an avatar icon replacement? Admittedly I didn't copy this post into my test earlier so didn't discover the problem, though that shall be easy to fix. However I'd like to confirm its usage so that I won't fix the wrong thing :smiley:

daattali:
Correct. I didn't want that image to be used as the big background image on the post page because that image is more rectangular, whereas the post image is a smaller square one

PR #503, opened 2019-06-05
Add permission for jekyll user on Gemfile in Docker

JackGruber:
Patch for docker build error: 
```
There was an error while trying to write to /srv/jekyll/Gemfile.lock. 
It is likely that you need to grant write permissions for that path.
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM jekyll/jekyll
 
-COPY Gemfile .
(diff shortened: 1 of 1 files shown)


npes87184:
Hi,

You can also revert 766b62238f38c47bbfd853b133c15e40e1acf002.

Thanks.

daattali:
Thank you to both of you

Issue #505, opened 2019-06-13
Consider removing gh-pages branch and instead recommending using the master branch even for project pages


daattali:
GitHub now allows the website to be built directly from the master branch, not only from `gh-pages`. For projects that are primarily a software project, and the code in their git repo is code for their software, it still makes sense to use `gh-pages` as their website branch. But since beautifuljekyll it meant to be a website, I think it makes sense for this repository to be served directly from `master` by default because it's not common that a beautifuljekyll fork needs separate `master` and `gh-pages` branch.

If anyone has good reasons against this, please let me know.

VincentTam:
> I think it makes sense for this repository to be served directly from `master` by default because it's not common that a beautifuljekyll fork needs separate `master` and `gh-pages` branch.

There's a situation that one separate these two branches.  In case of developping and contributing back to this theme, one might want to test the new features by merging them against the fork's `gh-pages` branch.  The `master` branch exists when one clicks the "Fork" button, and it serves as a base point for new branches.

However, there're usually much more users than developers, so perhaps the following line needs to be updated.

https://github.com/daattali/beautiful-jekyll/blob/94161dbef714a6ca0be122b46df3ea2b4e6c4064/staticman.yml#L28

daattali:
There are certainly cases where gh-pages makes sense, such as when this theme is used as the website of a code project (in which case `master` would host the actual project codebase and `gh-pages` would host beautifuljekyll).

But I think for the vast majority of users, `master` makes sense and gh-pages isn't necessary. And I think that for the few cases where gh-pages makes sense, it's for people who are advanced enough that they would know how to do that anyway. Would you agree?

VincentTam:
> But I think for the vast majority of users, `master` makes sense and gh-pages isn't necessary. And I think that for the few cases where gh-pages makes sense, it's for people who are advanced enough that they would know how to do that anyway. Would you agree?

Yeah, of course.  Thinking about some recent repos for Beautiful Jekyll sites using names other than `*.github.io`, I had omitted project page users capabilities.  So let's forget about my proposal.

abelcheung:
It looks like to me that, how users are integrating this theme into their workflow is a non-issue, under the context of this report. Be it a fork, file copy or whatever, people always do one thing in common: use the `master` branch of this theme as a basis. The `gh-pages` branch of this theme only serves as a distraction as it's unmaintained.

daattali:
[...]

PR #506, opened 2019-06-14
Abs link patch

abelcheung:
This is required fixes for absolute links. To use it, one would need to either:

1. Change `site.url` to new setting (remove project component), or
2. Simply delete `url` and `baseurl` altogether and let Github autoguess them

FAQ is not added yet, since I'm not sure how @daattali you would like
to express this issue inside FAQ. Suppose it can be added later?

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,10 +176,13 @@ If you're not sure what the difference is, you can probably safely ignore this s
 
 If you want to use this theme to host a website that will be available at `https://YOURUSERNAME.github.io`, then you do not need to read this section. That is called a User Page, you can only have one User Page in your GitHub account, and it is what you get by default when forking this project.
 
-If you want to use this theme to create a website for a particular repository, it will be available at `https://YOURUSERNAME.github.io/REPONAME`, and that is called a [Project Page](https://help.github.com/articles/user-organization-and-project-pages/). You can have a Project Page for each repository you have on GitHub. There are two important things to note when creating a project page:
+If you want to use this theme to create a website for a particular repository, it will be available at `https://YOURUSERNAME.github.io/REPONAME`, and that is called a [Project Page](https://help.github.com/articles/user-organization-and-project-pages/). You can have a Project Page for each repository you have on GitHub.
 
-1. In the configuration file (`_config.yml`), you should set `baseurl` to be `/projectname` instead of `""`.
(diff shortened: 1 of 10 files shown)


daattali:
Thanks @abelcheung this is great! I tested on a fork and it seems to work.

1. Would it not be safer to remove all uses of `relative_link` and replace them all with `absolute_link`?
2. Do we need to keep the url+baseurl params in the config file, do they have any use still?
3. This relates to the above question - when running this theme using ruby on a local server rather than via GitHub Pages, are the url+baseurl still not needed?
4. Let's modify the README section about "Creating a User Page vs a Project Page"

abelcheung:
1. I would still suggest treating relative / absolute links differently though, reasons below:
    1. Semantically it makes more sense (for html code inspection and debugging, for example)
    1. On situation where Jekyll doesn't trigger a site rebuild, keeping all links as absolute ones may break the site (toggling HTTPS support in Github for example?)
    1. Users may want different styling depending on link type (e.g. extra visual clue on external links)
2. For serving on Github, the settings can be removed altogether. But...
3. When serving on local servers, users still need to set them manually.
4. OK, will do soon.

daattali:
When serving locally, are both `url` and `baseurl` needed, or just `url`? Regardless of the answer, Since they're not needed anymore for github pages, these settings should move further down in the config file, so that the first settings are one that are more commonly changed.

I'll also have to make a new demo GIF when this PR is ready

abelcheung:
For local serving `baseurl` is always needed; it must be set to empty string (`baseurl = ""`) even when the site structure has no use of `baseurl`, otherwise Jekyll will complain and refuse to build the site.

About how much further down these settings would be moved, it's your call :wink:

daattali:
I made a few small changes to the config file https://github.com/daattali/beautiful-jekyll/pull/506/commits/d674dae9d836fa68cca6475a30b731c1987b40a4 can you verify that agree with the new text

daattali:
This seems very relevant: https://github.com/daattali/beautiful-jekyll/issues/505

Since this PR is already a breaking PR for existing projects that changes how project pages are served, it might be a good place to also mention gh-pages doesn't have to be used. Do you see any reason not to recommend using master?

abelcheung:
Because it's project page. :grin: Code and development stuff live in master branch, and project website is stored in `gh-pages` branch as an after thought. It is very rare for a project to create project website first before any code materializes; most likely they already have existing repository in use.

daattali:
OK I'll leave it for now. I understand what you're saying because project pages generally are meant to be a website of an actual project that has code, in which case it makes sense for the code to be in master and website to be in gh-pages. From my experience, most people I saw that create project pages actually use the website as the only code in their repo and don't actually have another more proper master. But I'll leave that for later. I'll look over the changes again soon and then merge!

abelcheung:
Thanks a lot for consideration. In the case your described, I think I can understand the usage pattern now (code and website living on separate repo). If necessary it's always possible to add such usage to FAQ later on :)

PR #508, opened 2019-06-23
Obsoletes navbarlink with relative_url filter

abelcheung:
The whole `navbarlink.html` is effectively a half baked version of `relative_url` filter (if URL scheme is found, return URL as-is, otherwise prepend `baseurl` to it).  One less file to maintain :grin:


daattali:
This is great! I wish that filter existed back in 2015, would have saved so much trouble. This is some good work towards https://github.com/daattali/beautiful-jekyll/issues/254 

By the way, I added you in the README as a major contributor

PR #509, opened 2019-06-25
Staticman: suggestion to use reCAPTCHA

VincentTam:
Due to inconvenience brought by massive spam comments reported in https://github.com/einsteinpy/einsteinpy.github.io/pull/25#issuecomment-504936235, perhaps we should at least give an advice on how to stop that.


daattali:
Thanks for staying on top of this!

PR #511, opened 2019-07-03
added ability to center images

JBrut22:
I added the ability to center images on pages and posts using markdown. I found it useful for one of my pages and thought others might think so too.

Apologies if you already had this functionality built-in, however, I could not find it in the markdown tutorial.

Can be performed by adding {: .center-image } at the end of an image line like so:
``` Image Name{: .center-image } ```

Thanks again for putting beautiful jekyll together! I found it extremely useful and I am grateful for it. Hopefully you will find this small addition useful as well.

### DIFF
--- a/css/main.css
+++ b/css/main.css
@@ -103,6 +103,11 @@ img {
   margin-top: -25px;
 }
 
+.center-image
+{
+    margin: 0 auto;
+    display: block;
+}
 /* --- Navbar --- */
 
 .navbar-custom {


abelcheung:
Looks like Bootstrap 3.3.x indeed has such facility included &ndash; [the `.center-block` class](https://getbootstrap.com/docs/3.3/css/#helper-classes-center).

daattali:
Thanks for both of you. I added an example in the sample post of how to use `.center-block`

PR #514, opened 2019-07-03
Staticman documentation enhancement

VincentTam:
# Webhook configuration

When `moderation: true` in `staticman.yml`, a PR containing a branch like `staticman_5525bf80-055f-11e9-a156-295f601b09e8` is created.  After merging such PR, such branch would become useless.  We have better offer an advice for automatic deletion after merging to avoid such branches from piling up.

The technical details consist of several steps.  I redirect interested users to the [relevant page in the official doc](https://staticman.net/docs/webhooks), and adding necessary adaptations.

# Changes in the comments in site config file
## A sample Staticman API endpoint

The current `endpoint` parameter is borrowed from `_includes/comments.html` in https://github.com/mmistakes/minimal-mistakes/pull/1845.  Due to [Staticman's URL scheme change introduced in v3](https://mmistakes.github.io/minimal-mistakes/docs/configuration/#staticman-v3), one has to add an extra parameter `service` which accepts either `github` or `gitlab` in the `<form action="url">`.  Unluckily, this has _not_ yet been documented in Staticman's official site.  As a result, some users are _not_ aware of such change, and spend time on fruitless try-and-error.  Different blog themes might have different logic to complete the `form action` URL.  For instance, in [Hugo Swift Theme](https://github.com/onweru/hugo-swift-theme#staticman-comments), in which I have been invited to collaborate, there's a `gitProvider` parameter.  Nonetheless, I've _no_ intention to change the template code in #440 to keep things working.  According to Staticman's author's comment in https://github.com/eduardoboucas/staticman/issues/264#issuecomment-460446511, Staticman's job is to get the HTML form data into the GitHub/GitLab repo, so the only appropriate place to document such theme-specific logic is the theme's README and relevant config files.

## Little complement of #509 

Sometimes, users might not read all of the section **Staticman comments** in the README.  Therefore, a short reminder in the comments might catch them.

# Updates in the comments in the repo config file

## Background [TL;DR]

I received a 500 error during comment submission to the public API instance for Framagit (i.e. https://staticman-frama.herokuapp.com/) in a private clone of this repo.  From the "server side", I know that I've got the correct API config parameters.  It took me hours of experiments to [rediscover the cause  of such 500 error](https://vincenttam.gitlab.io/post/2018-12-25-staticman-repo-setup-errors/): at the bottom of `staticman.yml`, under the section `reCAPTCHA`, I've kept the default `enabled: false` for convenience, but I left `secret` unchanged while changing the `endpoint` in `_config.yml`.  That reCAPTCHA `secret` was encrypted with this theme's default API instance (i.e. https://staticman3.herokuapp.com/), while the `endpoint` is holding _another_ private RSA key.  That would trigger a 500 error from the API server with a short and unhelpful response `{success: false}`.

## Steps for reproducing a 500 error under this theme

1. In `config.yml`: enable Staticman and change `endpoint` to *another* API instance.
2. Keep reCAPTCHA settings at the bottom of `staticman.yml` (i.e. reCAPTCHA disabled, but with non-empty `siteKey` and `secret`.)

Since the `endpoint` is changed, the corresponding RSA private key no longer matches the one used for encrypting reCAPTCHA `secret` in the `master` branch.  That would cause a 500 error from the API server, even if `enabled` is set to `false`, and reCAPTCHA parameters are empty in `_config.yml`.

## Proposed solutions
Either one will do.  @daattali please let me know which one you prefer.  The README might need further editing before it can get merged.
### Keep the reCAPTCHA section uncommented
In this case, no matter the user use reCAPTCHA or not, (s)he has to modify the repo config.

- enable reCAPTCHA: supply own `siteKey` and `secret`
- disable reCAPTCHA: change the values to `siteKey: ""` and `secret: ""`

I've written README based on this solution, but after creating this PR, I've found another simpler approach.

```yml
  # reCAPTCHA (OPTIONAL)
  # Register your domain at https://www.google.com/recaptcha/ and choose reCAPTCHA V2
  # Either use your OWN siteKey and secret, or change them to the empty string "".
  reCaptcha:
    enabled: false
    siteKey: "6Lcv8G8UAAAAAEqV1Y-XEPum00C_DxhD6O--qkFo"
    # (!) ENCRYPT reCaptcha secret key using Staticman /encrypt endpoint
    # i.e. https://staticman3.herokuapp.com/v3/encrypt/{your-site-secret}
    # For more information, https://staticman.net/docs/encryption
    secret: "p5uHlH9hCqpMJaGKXdt5MEWFo7K6fX8hoYUwR3aIafOI6rtItLauaDCkGOucysJtrVZy+sHffioGzMsOU64JFDSyPQgrXujegcOHFRXHhD4fOUuBXSvV+OZ8JhSPTGWaRcQcoiGX4pT5hlebLddOl59b6sn6kU1ODQcEbhP83xVLZlaTWOrNrF5Wvy3TMXpH5gyl1tZEORxADAShMYyUbNR7XZYLEg1DfgIBHfIg3cKwdFt7KVLejFGKIiBYRAZDE2JuHItNmzJ2x9JgSK3E+XnShV5tuWpncnyFonJVHGEky/zRfUVLHobDMcJ/u9nlZqE8u47W+833F1WaIYuwNw=="
```

### Comment out the whole reCAPTCHA section

- reCAPTCHA users can simply uncomment the parameters to get things work.
- Otherwise, one can simply leave this untouched.

I personally prefer this approach.  First, this is in line with the convention of switching off everything in this repo.  Second, this requires less editing for non reCAPTCHA users.  The only marginal drawback of this approach would be the additional difficulty to distinguish parameters from comments in the commented code below, but that shouldn't be an issue for normal users who know basic YML.  If you prefer this approach, please let me know, so that I'll edit the README accordingly.

```yml
  # reCAPTCHA (OPTIONAL)
  # Register your domain at https://www.google.com/recaptcha/ and choose reCAPTCHA V2
  # Either use your OWN siteKey and secret, or change them to the empty string "".
  # reCaptcha:
    # enabled: false
    # siteKey: "6Lcv8G8UAAAAAEqV1Y-XEPum00C_DxhD6O--qkFo"
    # (!) ENCRYPT reCaptcha secret key using Staticman /encrypt endpoint
    # i.e. https://staticman3.herokuapp.com/v3/encrypt/{your-site-secret}
    # For more information, https://staticman.net/docs/encryption
    # secret: "p5uHlH9hCqpMJaGKXdt5MEWFo7K6fX8hoYUwR3aIafOI6rtItLauaDCkGOucysJtrVZy+sHffioGzMsOU64JFDSyPQgrXujegcOHFRXHhD4fOUuBXSvV+OZ8JhSPTGWaRcQcoiGX4pT5hlebLddOl59b6sn6kU1ODQcEbhP83xVLZlaTWOrNrF5Wvy3TMXpH5gyl1tZEORxADAShMYyUbNR7XZYLEg1DfgIBHfIg3cKwdFt7KVLejFGKIiBYRAZDE2JuHItNmzJ2x9JgSK3E+XnShV5tuWpncnyFonJVHGEky/zRfUVLHobDMcJ/u9nlZqE8u47W+833F1WaIYuwNw=="
```

### DIFF
--- a/staticman.yml
+++ b/staticman.yml
@@ -101,8 +101,8 @@ comments:
   # Use your OWN siteKey and secret.
   reCaptcha:
     enabled: false
-    # siteKey: "6Lcv8G8UAAAAAEqV1Y-XEPum00C_DxhD6O--qkFo"
+    #siteKey: "6Lcv8G8UAAAAAEqV1Y-XEPum00C_DxhD6O--qkFo"
     # (!) ENCRYPT reCaptcha secret key using Staticman /encrypt endpoint
     # i.e. https://staticman3.herokuapp.com/v3/encrypt/{your-site-secret}
     # For more information, https://staticman.net/docs/encryption
-    # secret: "p5uHlH9hCqpMJaGKXdt5MEWFo7K6fX8hoYUwR3aIafOI6rtItLauaDCkGOucysJtrVZy+sHffioGzMsOU64JFDSyPQgrXujegcOHFRXHhD4fOUuBXSvV+OZ8JhSPTGWaRcQcoiGX4pT5hlebLddOl59b6sn6kU1ODQcEbhP83xVLZlaTWOrNrF5Wvy3TMXpH5gyl1tZEORxADAShMYyUbNR7XZYLEg1DfgIBHfIg3cKwdFt7KVLejFGKIiBYRAZDE2JuHItNmzJ2x9JgSK3E+XnShV5tuWpncnyFonJVHGEky/zRfUVLHobDMcJ/u9nlZqE8u47W+833F1WaIYuwNw=="
+    #secret: "p5uHlH9hCqpMJaGKXdt5MEWFo7K6fX8hoYUwR3aIafOI6rtItLauaDCkGOucysJtrVZy+sHffioGzMsOU64JFDSyPQgrXujegcOHFRXHhD4fOUuBXSvV+OZ8JhSPTGWaRcQcoiGX4pT5hlebLddOl59b6sn6kU1ODQcEbhP83xVLZlaTWOrNrF5Wvy3TMXpH5gyl1tZEORxADAShMYyUbNR7XZYLEg1DfgIBHfIg3cKwdFt7KVLejFGKIiBYRAZDE2JuHItNmzJ2x9JgSK3E+XnShV5tuWpncnyFonJVHGEky/zRfUVLHobDMcJ/u9nlZqE8u47W+833F1WaIYuwNw=="
--- a/staticman.yml
+++ b/staticman.yml
@@ -98,11 +98,11 @@ comments:
 
   # reCAPTCHA (OPTIONAL)
   # Register your domain at https://www.google.com/recaptcha/ and choose reCAPTCHA V2
-  # Either use your OWN siteKey and secret, or change them to the empty string "".
+  # Use your OWN siteKey and secret.
   reCaptcha:
     enabled: false
-    siteKey: "6Lcv8G8UAAAAAEqV1Y-XEPum00C_DxhD6O--qkFo"
+    # siteKey: "6Lcv8G8UAAAAAEqV1Y-XEPum00C_DxhD6O--qkFo"
     # (!) ENCRYPT reCaptcha secret key using Staticman /encrypt endpoint
     # i.e. https://staticman3.herokuapp.com/v3/encrypt/{your-site-secret}
     # For more information, https://staticman.net/docs/encryption
-    secret: "p5uHlH9hCqpMJaGKXdt5MEWFo7K6fX8hoYUwR3aIafOI6rtItLauaDCkGOucysJtrVZy+sHffioGzMsOU64JFDSyPQgrXujegcOHFRXHhD4fOUuBXSvV+OZ8JhSPTGWaRcQcoiGX4pT5hlebLddOl59b6sn6kU1ODQcEbhP83xVLZlaTWOrNrF5Wvy3TMXpH5gyl1tZEORxADAShMYyUbNR7XZYLEg1DfgIBHfIg3cKwdFt7KVLejFGKIiBYRAZDE2JuHItNmzJ2x9JgSK3E+XnShV5tuWpncnyFonJVHGEky/zRfUVLHobDMcJ/u9nlZqE8u47W+833F1WaIYuwNw=="
+    # secret: "p5uHlH9hCqpMJaGKXdt5MEWFo7K6fX8hoYUwR3aIafOI6rtItLauaDCkGOucysJtrVZy+sHffioGzMsOU64JFDSyPQgrXujegcOHFRXHhD4fOUuBXSvV+OZ8JhSPTGWaRcQcoiGX4pT5hlebLddOl59b6sn6kU1ODQcEbhP83xVLZlaTWOrNrF5Wvy3TMXpH5gyl1tZEORxADAShMYyUbNR7XZYLEg1DfgIBHfIg3cKwdFt7KVLejFGKIiBYRAZDE2JuHItNmzJ2x9JgSK3E+XnShV5tuWpncnyFonJVHGEky/zRfUVLHobDMcJ/u9nlZqE8u47W+833F1WaIYuwNw=="
--- a/staticman.yml
+++ b/staticman.yml
@@ -4,7 +4,8 @@
 # another one to handle posts.
 # To encrypt strings use the following endpoint:
 # https://{STATICMAN API INSTANCE}/v3/encrypt/{TEXT TO BE ENCRYPTED}
-# {STATICMAN API INSTANCE} defaults to staticman3.herokuapp.com
+# {STATICMAN API INSTANCE} should match the `endpoint` in the theme config
+# file. It defaults to "staticman3.herokuapp.com".
 
 comments:
   # (*) REQUIRED
@@ -22,8 +23,8 @@ comments:
 
   # (*) REQUIRED
   #
-  # Name of the branch being used. Must match the one sent in the URL of the
-  # request.
+  # Name of the branch being used. Must match the `branch` in the theme config
+  # file.
   branch: "master"   #  use "master" for user page
   #branch: "gh-pages" #  use "gh-pages" for project page
 
@@ -98,11 +99,12 @@ comments:
 
   # reCAPTCHA (OPTIONAL)
   # Register your domain at https://www.google.com/recaptcha/ and choose reCAPTCHA V2
-  # Use your OWN siteKey and secret.
+  # To enable reCAPTCHA, set `enabled` to "true", and uncomment `siteKey` and `secret`
+  # below, and use your OWN ones.
   reCaptcha:
     enabled: false
     #siteKey: "6Lcv8G8UAAAAAEqV1Y-XEPum00C_DxhD6O--qkFo"
     # (!) ENCRYPT reCaptcha secret key using Staticman /encrypt endpoint
-    # i.e. https://staticman3.herokuapp.com/v3/encrypt/{your-site-secret}
+    # i.e. https://{STATICMAN API INSTANCE}/v3/encrypt/{your-site-secret}
     # For more information, https://staticman.net/docs/encryption
     #secret: "p5uHlH9hCqpMJaGKXdt5MEWFo7K6fX8hoYUwR3aIafOI6rtItLauaDCkGOucysJtrVZy+sHffioGzMsOU64JFDSyPQgrXujegcOHFRXHhD4fOUuBXSvV+OZ8JhSPTGWaRcQcoiGX4pT5hlebLddOl59b6sn6kU1ODQcEbhP83xVLZlaTWOrNrF5Wvy3TMXpH5gyl1tZEORxADAShMYyUbNR7XZYLEg1DfgIBHfIg3cKwdFt7KVLejFGKIiBYRAZDE2JuHItNmzJ2x9JgSK3E+XnShV5tuWpncnyFonJVHGEky/zRfUVLHobDMcJ/u9nlZqE8u47W+833F1WaIYuwNw=="
--- a/staticman.yml
+++ b/staticman.yml
@@ -98,7 +98,7 @@ comments:
 
   # reCAPTCHA (OPTIONAL)
   # Register your domain at https://www.google.com/recaptcha/ and choose reCAPTCHA V2
-  # Use your OWN siteKey and secret.
+  # Either use your OWN siteKey and secret, or change them to the empty string "".
   reCaptcha:
     enabled: false
     siteKey: "6Lcv8G8UAAAAAEqV1Y-XEPum00C_DxhD6O--qkFo"
--- a/staticman.yml
+++ b/staticman.yml
@@ -25,8 +25,7 @@ comments:
   #
   # Name of the branch being used. Must match the `branch` in the theme config
   # file.
-  branch: "master"   #  use "master" for user page
-  #branch: "gh-pages" #  use "gh-pages" for project page
+  branch: "master"   #  use "master" for user page or "gh-pages" for project pages
 
   commitMessage: "New comment by {fields.name}"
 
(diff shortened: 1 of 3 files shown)


daattali:
I definitely prefer the second approach too. But what's the point of the `enabled: false` parameter under reCaptacha, I thought that should serve exactly this purpose, of turning recpatcha on/off. Can't we make it so that if enabled:false then everything else is ignored?

VincentTam:
> I definitely prefer the second approach too. But what's the point of the `enabled: false` parameter under reCaptacha, I thought that should serve exactly this purpose, of turning recpatcha on/off. Can't we make it so that if enabled:false then everything else is ignored?

Thanks for thoughtful feedback!  My mind had blocked me.  In between the two, there's a better way.  I've just tested it at with [`289412b..425de1c`](https://github.com/VincentTam/beautiful-jekyll/compare/289412b..425de1c) on my demo site: https://git.io/bjsm18.  From https://github.com/VincentTam/beautiful-jekyll/pull/24 created just after that, it's clear that your suggestion works.

```yml
  # reCAPTCHA (OPTIONAL)
  # Register your domain at https://www.google.com/recaptcha/ and choose reCAPTCHA V2
  # Use your OWN siteKey and secret
  reCaptcha:
    enabled: false
    # siteKey: "6Lcv8G8UAAAAAEqV1Y-XEPum00C_DxhD6O--qkFo"
    # (!) ENCRYPT reCaptcha secret key using Staticman /encrypt endpoint
    # i.e. https://staticman3.herokuapp.com/v3/encrypt/{your-site-secret}
    # For more information, https://staticman.net/docs/encryption
    # secret: "p5uHlH9hCqpMJaGKXdt5MEWFo7K6fX8hoYUwR3aIafOI6rtItLauaDCkGOucysJtrVZy+sHffioGzMsOU64JFDSyPQgrXujegcOHFRXHhD4fOUuBXSvV+OZ8JhSPTGWaRcQcoiGX4pT5hlebLddOl59b6sn6kU1ODQcEbhP83xVLZlaTWOrNrF5Wvy3TMXpH5gyl1tZEORxADAShMYyUbNR7XZYLEg1DfgIBHfIg3cKwdFt7KVLejFGKIiBYRAZDE2JuHItNmzJ2x9JgSK3E+XnShV5tuWpncnyFonJVHGEky/zRfUVLHobDMcJ/u9nlZqE8u47W+833F1WaIYuwNw=="
```

I'm going to update README now.

daattali:
Much cleaner! I would say that the parameters don't even need to be commented out though. Nothing will happen if they're left uncommented. Only params that must be ignored should be commented out

VincentTam:
> Much cleaner! I would say that the parameters don't even need to be commented out though. Nothing will happen if they're left uncommented. Only params that must be ignored should be commented out

No, even if `enabled: false`, if `secret` are either null or not encrypted with the RSA private key of the API instance defined in `endpoint`, then the Staticman API server will throw a 500 error.  That's why there's a need to comment out `recaptcha.siteKey` and `recaptcha.secret`, due to the convention to disable everything by default.  Non-users of reCAPTCHA shouldn't get interfered with `recaptcha.siteKey` when they adopts another API `endpoint`.

VincentTam:
To my observations more convincing, I've made this experiment.

1. Cloned this project (GitHub allows at most one fork): https://github.com/VincentTam/BJPubTest at 2f4147c.
2. Did minimal config at https://github.com/VincentTam/BJPubTest/commit/042c7cf87af21c4ba265058423309fb4667faef4 to set up GitHub project pages.
    - **Changed `endpoint`** to the [public Staticman API instance for Framagit](https://framagit.org/staticmanlab1):

          https://staticman-frama.herokuapp.com/v3/connect/github/VincentTam/BJPubTest
    - Invited @staticmanlab.
    - Adjusted the `branch` and `repository`.
3. Browsed the top post in the generated site: https://vincenttam.github.io/BJPubTest/2015-02-28-test-markdown/.
4. Left a sample comment.  Observed a 500 error.
Screenshot from 2019-07-05 18-32-15
5. Created a clone of this clone: https://github.com/VincentTam/BJPubTest1.
6. Commented out the unused reCAPTCHA parameters at https://github.com/VincentTam/BJPubTest1/commit/81627696e87350db54efbcf59b08325fa591c83e and invited @staticmanlab again.

       https://staticman-frama.herokuapp.com/v3/connect/github/VincentTam/BJPubTest1
7. Browsed the top post in another generated site: https://vincenttam.github.io/BJPubTest1/2015-02-28-test-markdown/.
8. Input the *same* text again.  The *same* API server gives a 200 response, indicating a success.
Screenshot from 2019-07-05 18-55-04
9. Since `moderation` is disabled by default, the comment was pushed directly to the repo at https://github.com/VincentTam/BJPubTest1/commit/8fd650873e3fac6c1149f03a1ca426535a19e9a3.
https://github.com/daattali/beautiful-jekyll/blob/2f4147c00836ada93dad596f721b87274306931b/staticman.yml#L51-L54

Since "BJPubTest1" is a clone of "BJPubTest", they are identical except these two parameters.

| `repository` in `_config.yml` | `VincentTam/BJPubTest` | `VincentTam/BJPubTest1` |
| --- | --- | --- |
| `recaptcha.secret` in `staticman.yml` | uncommented | commented |
| result | failed | success |

The room for improvement in my previous commit on Staticman repo config file is the whitespace on the right of the leftmost `#`.
https://github.com/daattali/beautiful-jekyll/blob/1d1833fcd071ba040959a367111405fda385bc4c/staticman.yml#L102-L108
This can be deleted so as to be consistent with
https://github.com/daattali/beautiful-jekyll/blob/1d1833fcd071ba040959a367111405fda385bc4c/staticman.yml#L27-L28

daattali:
Sorry if this is completely wrong, but I have to ask because it doesn't make sense to me. In my mind, if recaptcha is disabled, then the parameters for recaptcha shouldn't even be passed along - they should just be ignored. So I'm sorry if this question doesn't make sense, but:

What if in here, we add a check to only include these 2 parameters **if** recaptcha is enabled?
https://github.com/daattali/beautiful-jekyll/blob/d1c8b2fee19fe600ad6b08791b270d0aa04839a3/_includes/staticman-comments.html#L46

VincentTam:
> What if in here, we add a check to only include these 2 parameters if recaptcha is enabled?
> https://github.com/daattali/beautiful-jekyll/blob/d1c8b2fee19fe600ad6b08791b270d0aa04839a3/_includes/staticman-comments.html#L46

You suggested modification _won't_ change the result, because in BJPubTest (failed site)'s [`_config.yml`](https://github.com/VincentTam/BJPubTest/blob/2f4147c00836ada93dad596f721b87274306931b/_config.yml#L137-L138), the two reCAPTCHA parameters are `null`.

```yml
# BJPubTest's _config.yml
  reCaptcha:
    siteKey  : # Use your own site key, you need to apply for one on Google
    secret   : # ENCRYPT your password by going to https://staticman3.herokuapp.com/v3/encrypt/<your-site-secret>
```

As a result,  
Screenshot from 2019-07-06 09-32-59  
in the failed example, the hidden fields _doesn't_ contain `options[reCaptcha][siteKey]`.  
Screenshot from 2019-07-06 09-47-21

> Sorry if this is completely wrong, but I have to ask because it doesn't make sense to me. In my mind, if recaptcha is disabled, then the parameters for recaptcha shouldn't even be passed along - they should just be ignored. So I'm sorry if this question doesn't make sense, but:

To make sense of my experiment, start thinking from the only factors changed in the two setups:

1. `repository` in site config: necessary for sending HTTP POST requests to the right places.
2. whether the two unused reCATPCHA comments are commented out in `staticman.yml`

Results: see the above table.

`staticman.yml` is Staticman config, which _doesn't_ (and _shouldn't_) care whether the GitHub/GitLab repo uses GitHub/GitLab Pages or not, according to Staticman's author's comment in https://github.com/eduardoboucas/staticman/issues/264#issuecomment-460446511.   This config file is used by the API instance for communications between Staticman API server and GitHub/GitLab server.  In the failed example, the two unused reCAPTHCA parameters are loaded; in the successful one, they aren't loaded.

Why your suggestion won't change the results?  
As long as the input HTML form data structure matches the requirements in this repo config file, then Staticman will perform its work.  As can be seen from the successful example (BJPubTest1), the HTTP POST request has the correct structure, so the API responsed correctly.  Commit https://github.com/VincentTam/BJPubTest1/commit/81627696e87350db54efbcf59b08325fa591c83e shows that the failed example (BJPubTest) is using exactly the *same* theme template, so the input HTML form data should have the correct structure.

What is the possible explanation?  
The loaded unused parameters interfere the server-side communications even if reCAPTCHA is disabled in both `_config.yml` and `staticman.yml`.

My conclusion from this experiment is that the lack of `#` caused this error *in my Framagit instance* for the message "Hello world" with `siteKey` and `secret` unchanged.  You're free to feel dubious about this, because you don't get such error during your testing with https://github.com/daattalitest/daattalitest.github.io.  My explanation for your test results is that the default paramaters were tested with the default Staticman API instance, which holds a _different_ RSA private key that was used for the encryption of the reCAPTCHA's site secret.  You can test it with other Staticman instances by searching [users "staticman" on GitHub](https://github.com/search?q=staticman&type=Users).  Another way to verify my conclustion is to break the reCAPTCHA site secret with strings like "123" while keeping `enabled: false`.

daattali:
Thanks for all the detailed writeup explanations.

So if we go with the current minimal changes, would the user have to uncomment siteKey and secret when they set enabled:true ? If so, that should be mentioned in the yml file

staticman.yml
```diff
@@ -4,7 +4,8 @@
 # another one to handle posts.
 # To encrypt strings use the following endpoint:
 # https://{STATICMAN API INSTANCE}/v3/encrypt/{TEXT TO BE ENCRYPTED}
-# {STATICMAN API INSTANCE} defaults to staticman3.herokuapp.com
+# {STATICMAN API INSTANCE} should match the `endpoint` in the theme config
+# file. It defaults to "staticman3.herokuapp.com".
```
VincentTam: More detailed explanations to prevent setup failure.

staticman.yml
```diff
@@ -22,8 +23,8 @@ comments:
 
   # (*) REQUIRED
   #
-  # Name of the branch being used. Must match the one sent in the URL of the
-  # request.
+  # Name of the branch being used. Must match the `branch` in the theme config
+  # file.
```
VincentTam: User's won't actually see the POST URL without digging into the code.  Referring to `theme` in `_config.yml` is more meaningful.

staticman.yml
```diff
@@ -98,11 +99,12 @@ comments:
 
   # reCAPTCHA (OPTIONAL)
   # Register your domain at https://www.google.com/recaptcha/ and choose reCAPTCHA V2
-  # Use your OWN siteKey and secret.
+  # To enable reCAPTCHA, set `enabled` to "true", and uncomment `siteKey` and `secret`
+  # below, and use your OWN ones.
```
VincentTam: More detailed instructions due to previous comment.

staticman.yml
```diff
@@ -98,11 +99,12 @@ comments:
...
   reCaptcha:
     enabled: false
     #siteKey: "6Lcv8G8UAAAAAEqV1Y-XEPum00C_DxhD6O--qkFo"
     # (!) ENCRYPT reCaptcha secret key using Staticman /encrypt endpoint
-    # i.e. https://staticman3.herokuapp.com/v3/encrypt/{your-site-secret}
+    # i.e. https://{STATICMAN API INSTANCE}/v3/encrypt/{your-site-secret}
```
VincentTam: Allow different Staticman bot, like @staticmanaima, @Bloggerbust-bot, etc.

daattali:
The sample siteKey and secret - should they ever be used? Or are they only for demo purposes?

VincentTam:
> The sample siteKey and secret - should they ever be used? Or are they only for demo purposes?

These are *my own siteKey and secret*, so for others, they're only for display.  That's why I've emphasized "OWN" in the Staticman config file.

List of domains hooked up with my reCAPTCHA v2 site key  
Figure: List of domains hooked up with my reCAPTCHA v2 site key

daattali:
Thanks Vincent

Issue #117, opened 2016-10-10
Few issues with code blocks, and some questions. 


darkhax:
After testing out a this theme, I have noticed a few issues with code blocks. (not that I am using ' rather than ` for this issue, because I couldn't work out how to escape it lol) I get the feeling that these issues are more so limitations of the tools used rather than this project, so at the very least this will help others who are having these issues. 

The first issue is related to syntax highlighting. I believe there may be some case sensitivity with it. If you were to use '''Java it will not work. But if you use '''java it will. 

The second issue is with new lines. If you start a code block directly after text, it will break the formatting. To prevent this, a new line must be between them. 

As for my questions, I am wondering if line wrapping on code blocks has been worked out. I am currently writing tutorials about Java programming, and the line wrapping on code blocks can be quite the eye sore. You can see an example page [here](http://tutorials.darkhax.net/tutorials/custom_entity_data/). I had tried fixing it myself, however the best I could do is disable all line wrapping, which turned all code blocks into one line. I am not very familiar with css though lol. 

I have also noticed that the url in the top left links to the github site rather than my domain, which throws a 404. Would you happen to know what I did wrong? The repo for the site can be found [here](https://github.com/darkhax/tutorials.darkhax.net)

Thanks for your time, and this awesome theme :)

daattali:
Hi @darkhax , as you suspected, all these issues are 100% to do with markdown, and I can't offer much help with them.  I also run into random issues like these often with markdown (especially with code blocks, they really are trickier), and every time I just do a lot of googling. Unfortunately markdown code blocks are not super trivial like most other markdown features.  The issue with the new line is pretty standard though with markdown - it's the same problem that lots of people have where they can't figure why you always need to have a blank line (ie press Enter **twice**) in order to actually have a blank line in the output.  All these things are just markdown difficulties that we all have to do deal with... Sorry I can't be of more help.  It's good that these problems are happening mostly with code blocks and not with other formatting elements, because generally if you're writing code blocks that means you have some basic coding knowledge and should be able to figure out how to fix these issues :)

daattali:
One thing you can experiment with: there are different markdown parsers and different syntax highlighters.  They all work differently and each have their own issues. You can change them here: https://github.com/daattali/beautiful-jekyll/blob/master/_config.yml#L101-L102

You can read the jekyll documentation to see what other markdown/highlighter options are available. I don't remember the difference between them or why I chose these specific ones, but you can experiment with them.

[...]

PR #516, opened 2019-07-06
Refactor code block styles

### LINKED ISSUE #512: Making code blocks in line with custom page color
Recently I'm working on some changes to simplify code block styles and allow them to blend in more nicely with custom page background color (through more extensive use of transparency). Would such changes be welcome, or @daattali you would prefer styles staying as-is? About elements with hardcoded color, tables and code blocks come into mind, there are probably other as well. In the case of code block, I mean something like the following instead of hardcoding gray shades in code block background:

Clipboard02

image

image

abelcheung:
Code block bg adopts to custom user bg color instead of using
hardcoded color, preserving original design as much as possible ( Closes #512 )

------
Some background info:

- White space related change described in #117 is already available from bootstrap, so I removed it
- Lineno related styles moved to main css

### DIFF
--- a/css/main.css
+++ b/css/main.css
@@ -674,8 +674,9 @@ pre {
   margin: 0;
 }
 .highlight > pre {
-  background-image: repeating-linear-gradient(
-    rgba(0,0,0,0.03), rgba(0,0,0,0.03) 1.5em, rgba(0,0,0,0.06) 1.5em, rgba(0,0,0,0.06) 3em);
+  background-image: linear-gradient(
+    rgba(0,0,0,0.06), rgba(0,0,0,0.06) 1.5em, rgba(0,0,0,0.03) 1.5em, rgba(0,0,0,0.03) 3em);
+  background-size: auto 3em;
   background-position-y: 10px;
   border: 1px solid rgba(0,0,0,0.1);
   border-left: 7px solid #444;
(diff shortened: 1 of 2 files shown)


daattali:
What's the browser support for this?

abelcheung:
All current major ones I can think of running on Windows 7, even including IE 11. Though for Safari I can't be absolutely sure yet as I don't have mac and iphone.

daattali:
I just tested this out. Looks great! Only one very small thing I noticed when comparing old to new: The half-life padding on top is the same colour as the first line of code, which results in a taller area with the same colour.

Demo:

Old look
image

New look
image

I don't mind the actual colour being different , but notice that previously there was some different coloured padding above the first line

abelcheung:
Ah, I was careless about the strip order, where darker strip should apply on every first line instead of every second line. Shouldn't be hard to fix, will attend to it later today.

abelcheung:
Pushed additional commit to fix the stripping. About the change of color, the original reason is that the stripping became virtually undistinguishable with relatively darker background, so I added extra opacity to make them more visible. Right now the strips are set to 3% and 6% opacity; if it is desired to achieve the same pixel value as before on white bg, setting them to 1% and 2% respectively would suffice.

daattali:
Thanks!

Issue #518, opened 2019-07-09
[Proposal] Simplify css and js options?


abelcheung:
Currently there are multiple variable names associated with CSS usage:

- `common-css` and `common-ext-css` under layout
- `css` and `ext-css` under page
- effectively `layout.common-googlefonts` and `page.googlefonts` are also CSS, only saving a few keystrokes

Would it be useful to just unify them altogether and have a single `css` variable exposed under layouts and pages?

The same applies to javascript too, not just CSS. But javascript may need extra consideration though, as dependency problem and double inclusion problem affects user experience more than CSS do. As currently hand-coded loading order is used, it might be necessary to educate users not to mess with the order without careful consideration. :man_shrugging:

daattali:
Unifying sounds like a good idea! As long as users can still specify both common files that are used on all pages, and specific files that are used on individual pages

abelcheung:
Asking for comment before CSS changes finalise. CSS related settings would be split across `layout.common-css`, `layout.css` and `page.css`. Turns out it's not quite possible to unify into single variable, since some styles are global across whole site, and some others only apply to specific layout. It could be worked around by specifying site-wide styles inside `_config.yml`, but that's quite inconvenient for local development (any global config change needs a restart of `jekyll` everytime).

From users' perspective, sample config for layout and pages look like the following for now:

### `_layouts/base.html`
```yaml
---
common-css:
  - "/css/bootstrap.min.css"
  - "https://fonts.googleapis.com/css?family=Lora:400,700,400italic,700italic|Open+Sans:300italic,400italic,600italic,700italic,800italic,300,400,600,700,800"
  - "https://maxcdn.bootstrapcdn.com/font-awesome/4.6.0/css/font-awesome.min.css"
  - "/css/main.css"
---
```

### `_layouts/page.html` and `_layouts/post.html`
```yaml
---
css:
  - "/css/bootstrap-social.css"
---
```

### Individual CSS per post under `_post/`
```yaml
---
css:
  - "/css/myspecial.css"
---
```

----

One point still unsure of: whether an entry point should be provided for user to override any style, or simply tell users to change `main.css` directly.

abelcheung:
@VincentTam Wish to ask, would you mind if I move Staticman related CSS and JS around, so they become more integrated into the reorganisation idea above? I'm working on CSS for now, javascript changes would be pending later.

VincentTam:
[...]

PR #520, opened 2019-07-12
Css var simplify

abelcheung:
CSS part of the simplification mentioned in #518. Don't merge yet, user customisation stylesheet support will be added in coming patch, and site config may need some more comment.


abelcheung:
need thinking over impact of change b4 resubmitting

daattali:
How about a more simplified PR that just merges the `ext-` css files with the local css files? That would already be an improvement

PR #522, opened 2019-07-23
Fix JavaScript loading order

### LINKED ISSUE #342: JS loading order problem
I've noticed a problem caused by the loading order of the JS files

## Scenario
- Add a custom Javascript file in a new blog post or page by:
  - using the YAML key js
  - writing your Javascript directly in the file within the `<script>` tag
- If you're using jQuery or other frameworks which were defined in the layout the execution fails

## Analysis

The include snipped `footer-scripts.hml` adds the `js` files in the wrong order. It adds the custom files first instead of the layout and global wide files. This results into an situation where you can't use jQuery, Ajax or Bootstrap related functions because the were loaded afterwards...

danielmpetrov:
Fixes #342

This request simply changes the order scripts are loaded in footer-scripts.html, which addresses an issue where jquery and boostrap are undefined in page scripts.

The loading order proposed is **common-ext-js -> common-js -> ext-js -> page-js** because common scripts should be loaded before page scripts, and external scripts should be loaded before local scripts for the same reason, which is that local scripts may want to reference APIs from external sources.

### DIFF
--- a/_includes/footer-scripts.html
+++ b/_includes/footer-scripts.html
@@ -4,18 +4,6 @@
   {% endfor %}
 {% endif %}
 
-{% if page.ext-js %}
-  {% for js in page.ext-js %}
-    {% include ext-js.html js=js %}
-  {% endfor %}
-{% endif %}
-
-{% if page.js %}
-  {% for js in page.js %}
-    <script src="{{ js | relative_url }}"></script>
-  {% endfor %}
-{% endif %}
-
 {% if layout.common-js %}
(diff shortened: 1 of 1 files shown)


daattali:
Thanks!

@OCram85 is this what you had in mind?

daattali:
Pinging @OCram85 again :)

OCram85:
Sorry guys, I was on vacation in Scotland the last weeks :D

Looks fine for me 👍

PR #533, opened 2019-08-14
Add support for customizable date format

bangjunyoung:
The title says all. :-)


daattali:
Good addition, thanks!

PR #577, opened 2020-01-27
Fix footer does not stick to the bottom

### LINKED ISSUE #576: Footer stick to the bottom when contents are short or absent
I know this has been asked already, but it really feels like an issue on large screens or when the content is just a few lines.

The footer does not stay at the bottom of the page but immediately below the content, and the rest of the page shows the background.

Contact

If it may be of any help I'm willing to open a PR or try to fix it on my own. I've tried several solutions but with no success. I'll keep trying though, I wonder if you have any suggestions on how to approach the problem.

I don't want to modify the current behavior of the footer, but just make it at the bottom of the page for those that don't have many lines. I've checked solutions like [this](https://jsfiddle.net/osvx1zoo/3/) but they are not valid options.

simoarpe:
## Resolves #576 

This PR does not affect the current behavior of the footer except for the edge case described in the issue: when there is a small content or none the footer now stays at the bottom the page.

Because the height of the footer can change, most of the solutions around didn't work. Using `flexbox` for a sticky footer not only doesn't require any extra elements, but allows for a variable height footer.

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -20,13 +20,13 @@ common-js:
   {% include head.html %}
 
   <body>
-
+    <div class="content">
     {% include gtm_body.html %}
   
     {% include nav.html %}
 
     {{ content }}
-
+    </div>  
     {% include footer.html %}
   
     {% include footer-scripts.html %}
--- a/css/main.css
+++ b/css/main.css
@@ -6,8 +6,13 @@ layout: null
 
 /* --- General --- */
 
+html, body {
+  height: 100%;
+}
 body {
   font-family: 'Lora', 'Times New Roman', serif;
+  display: flex;
+  flex-direction: column;
   font-size: 18px;
   color: #404040;
   position: relative;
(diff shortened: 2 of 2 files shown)


daattali:
Thanks for the PR! 

This changes the layout of the entire page. Since this theme is used by tens of thousands of people, I'm extremely cautious in making such changes. I would like to see this tested with different config parameters, to make sure it doesn't break any existing features. How much have you tested this outside of your own project?

simoarpe:
I ran some tests, but probably not as much as it deserved.
Tried on different browsers, desktop and mobile.

Mutmatt:
Might be cautious of flex-direction with FF https://caniuse.com/#search=flex-direction

daattali:
Thank you @Mutmatt 

Have you tried this solution to see what it looks like on FF? I just rewrote the majority of the HTML and CSS while upgrading to bootstrap 4, so now seems like a good time to test this as well.

daattali:
@simoarpe the theme recently underwent huge changes, updating to bootstrap 4 which uses flexbox, so if you could update this PR to work with the current codebase, I'd be happy to accept this

daattali:
Closing this PR due to inactivity. If anyone wants to revive it in the future, feel free to submit a new one!

Issue #339, opened 2018-04-02
Is remote_theme supported?


chfast:
I tried to use

```
remote_theme: "daattali/beautiful-jekyll"
```

in my GitHub Pages config. This does not seem to produce good results (https://chfast.github.io/How-to-explicit-NOT-initialize-variables-in-cpp/). 

Should that work?

daattali:
This is a fairly new feature of github (allowing custom themes to use `remote_theme`), and I haven't yet found the time to make beautiful-jekyll compatible with that. So it does not work yet.

If someone wants to help making the necessary changes, I would appreciate the help!

dkoprowski:
Hi @daattali - some news on that :) ?

daattali:
Sorry @dkoprowski I still haven't looked! If you're familiar with it, I would not oppose to a pull request if you'd like to help :)

dkoprowski:
Unfortunately I'm not familiar with Ruby and Gems ;)

abelcheung:
Some great (and bad) news: I have managed to using it as `remote_theme` in Github Page (running in [one of my pet project](https://abelcheung.github.io/rifiuti2/) right now, thanks for the great work!), though quite a lot of files have to be cloned manually in order to get it to work.

With my (limited) understand of the `remote_theme` feature, the following 4 folders can be utilized without copying:
* `_layouts`
* `_includes`
* `_sass` (probably can be ignored since this theme is not using sass)
* `assets`

So the major obstacle probably involves moving css, javascript, images and all sort of include files to their proper place, and update links / references to them. I'm still new to Jekyll and even Ruby in general, so perhaps somebody more experienced can correct me if anything is missing. :)

abelcheung:
For those interested, one can check out [my example repo](https://github.com/abelcheung/site-test/) which is using a restructured beautiful-jekyll as remote theme. At a minimum, only a few config files and top level pages need to be copied in order to establish a working site. In particular, notice addition of `plugin` entries and `remote_theme` setting in `_config.yml`.

On second thought, I doubt this feature goes against daattali's interest, since the essence of `remote_theme` is to avoid forking and only copy files necessary for overriding defaults, thus not included into fork count (as an easily monitored user count of this theme). Unless daattali gives the green light and users do really want the feature, perhaps pull request won't be submitted.

daattali:
[...]

Issue #269, opened 2017-10-19
combine gem branch into master branch


daattali:
so that I won't have to support development in parallel branches that are nearly identical

gpotter2:
Nowadays with `remote_theme` supported natively by github pages, if you made a theme that followed https://jekyllrb.com/docs/themes/ it would be plug and play.

I've tested it: if you move the `css/` and `js/` folders (all non default `_something` folders) to an `assets/` folder, it works just fine.

daattali:
That's great! I don't know anything about ruby and haven't looked into github remote  themes, so I would appreciate if you wanted to make a pull request.

daattali:
@gpotter2 if you'd like to make a PR for this, it should be much smoother than the previous one :)

daattali:
I've released a new gem and combined it with master. I'll add documentation soon and delete the gem branch.

PR #590, opened 2020-03-24
Make compatible with remote_theme

gpotter2:
Add support for `remote_theme`:
- move all css/js/img to `assets`
- update all references accordingly
- make all data files optional
- add default config values
- doc

This is the "easy", low effort way of making it work. Eventually you could use scss...

@daattali 
fix https://github.com/daattali/beautiful-jekyll/issues/339


gpotter2:
Small demo of a website that uses remote-theme:
https://github.com/gpotter2/secdev.github.io
https://gpotter2.github.io/secdev.github.io/

skalee:
Data files are not available. This breaks social icons in the footer, for instance. Can be workarounded by copying these files to the target repo.

daattali:
@skalee is there no way to fix this without requiring more work by the end user?

gpotter2:
We can probably move it to assets to

skalee:
@daattali No idea. I'm a Jekyll beginner.

----

Also, some parts of theme's `_config.yml` need to be copied to the project's `_config.yml` as well. In particular, this applies to following colour definitions:

https://github.com/daattali/beautiful-jekyll/blob/a04be276e39bd3fc7acde9c785ae278bbad443c7/_config.yml#L39-L49

Otherwise, blank strings are used, and resulting CSS is slightly broken, for example this rule:

https://github.com/daattali/beautiful-jekyll/blob/a04be276e39bd3fc7acde9c785ae278bbad443c7/css/main.css#L73-L77

becomes:

```css
::selection {
  color: white;
  text-shadow: none;
  background-color: ;
}
```

In the result, selected text is white font on white background, at least in my browser.

----

Having said that, remote-theme capability is a HUGE improvement, even if a few additional steps must be done by user.

daattali:
Thanks both

This is indeed something I procrastinated with for way too long. I just spent some time reading over the documentation for this to understand it better.

I'm 100% onboard with merging this PR! But I do want to work out the little quirks so that it'll be a smooth experience for the user. It seems like there are two issues currently:

1. The `_data` folder doesn't get used in a gem-based theme. Do you @skalee

daattali:
Also, would merging this PR mean that the `master` branch would be a proper ruby gem and https://github.com/daattali/beautiful-jekyll/issues/269 would be solved organically, correct?

daattali:
I'd be very happy to get this PR merged if you get some time to continue the discussion

gpotter2:
I think the user will need to copy `_data` anyways. Assets isn't able to provide its data and it seems adding "default _data" is on the roadmap for jekyll 4.1.

- no this isn't enough to make beautiful-jekyll a gem theme (you'll need a .gemspec....)

gpotter2:
@daattali This PR will not break anything: forking the repo will still work. However it will enable us to use remote theme by only providing `_data`. Please merge asap

daattali:
Thanks for letting me know about the 4.1 Jekyll roadmap - it does look like soon enough the `_data` issue won't be an issue anymore.

I do want to merge this PR ASAP, but only when it is complete. I know it won't **break** anything, but I don't want to have a feature that's not fully working.

From my current understanding all that needs to happen to make this PR work 100% is:

1. Add a couple sentences in the README about how to use this theme as a `remote` theme. This should include instructions for copying the `_data` folder!

2. Any CSS values, or any other values on the site, that **require** config values from the config file should use some sensible default value, to accommodate the case of when the config file is empty

gpotter2:
I've added some doc. Feel free to review it.
I'm not up for replacing all the values with default ones..

Note this PR won't break the classic "clone" install anyways :smile: the only one currently deployed.

daattali:
In the README the first step is " have a working jekyll project" - do you mean having a jekyll project locally? Or do you mean just have a github repository? I was hoping this would work with github pages, and the documentation should make it clear how to use this on a GH repo

I completely understand you not wanting to deal with the default values issue - I'm doing a lot of work for free these days analyzing COVID so I don't have much free time to dedicate to this, hopefully someone else can find some time to do that

gpotter2:
I've looked into it:
- staticman uses `_data/ui-text`, but already has default values in place. No changes are required
- the footer uses `_data/SocialNetworks.yml`, only to get the icons & link templates.

The thing is that there's no real default value that would work for the footer links. Should we disable the links entirely when the file isn't available?
If so, should we also move the social network configs from `_config.yml` to `SocialNetworks.yml` for consistency? Otherwise people wouldn't realize it's not being displayed when the `SocialNetworks.yml` is missing.

I'll try to work on this @daattali. Thanks for your help
Stay safe

daattali:
@gpotter2 Could you submit commit dcd16a0391fb7972d1ae70d2fc73a955c315ed12 (where you add default values) as a separate PR and reference https://github.com/daattali/beautiful-jekyll/issues/600 ? 

If you don't want to do that extra work, let me know and I can do it later.

I'll take a look at the rest of what you did/said when I get a chance. Thanks!

daattali:
The "Add support for round avatar" commit should also be a separate commit as it's a completely separate issue, but that one I would also merge immediately

daattali:
@gpotter2 (I deleted the previous comments in favour of putting all the info here)

1. You created 3 commits, the first one is specifically relevant to this PR, the second one (adding default values) is related to this PR but can also be standalone, and the third one (round avatar) is not related to the PR. Do you mind submitting the last 2 commits as separate PRs? The default values commit should make a reference to #600 . This PR shouldn't have those two commits in it.

2. According to your documentation, does the remote_theme not work with github pages? Will it only work in local ruby installations? From my (limited) understanding, I thought remote_theme is supported by github pages and should be very similar to how it works with local installations

3. Regarding staticman: we can just add a single sentence in the readme under the remote_theme section that says if you're using staticman, you should copy the `_data/ui-text` folder

4. Regarding social network links: because in the near future this problem will be solved automatically (when jekyll includes the `_data` folder in the next version), I would say that we shouldn't touch the code too much - leave everything related to this feature where it currently is. Add a simple check in the code that checks if this data is available and if it isn't then just don't try to render this section. In the README under the remote_theme section, add a sentence that says that you **need to** copy the `_data/SocialNetworks.yml` folder

Does this all make sense?
Thanks for all your work

gpotter2:
Hi,

1. Makes sense, will do
2. Yes the goal is first and foremost to add support for remote-theme. I'm using https://github.com/gpotter2/secdev.github.io/blob/gh-pages/_config.yml as test platform. Build is currently failing but I'll fix it soon enough
3. ok
4. I prefer my changes because i don't believe users ever read the doc (e.g. the massive amount of PRs you get that are mistakes, despite the template). If i were you i would make it so that it either works, either not at all. Anything in between will bring you even more issues of confused users...

daattali:
Sorry I'm confused by the answer to 2. "Yes" means that GitHub Pages is supported, or not?

I agree that people won't really read the full docs. What do you think of having a sentence inside the config file , above the start of the social networks list, that says "If you're using remote_theme, you need to copy the xxx folder for these to work"

gpotter2:
Github pages is supported through `remote_theme`

daattali:
Great! The first goal of this theme is to be GitHub Pages-first, so I would prefer to say in the readme how to do this in a GitHub Pages site (like how most of the documentation focuses on github pages) rather than saying "create a jekyll project".

gpotter2:
@daattali I think I've addressed all your points.
I also squashed the commits when necessary

gpotter2:
Also note you shouldn't include `Gemfile.lock` in a gem. see https://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/

I've "only" updated it here because I assume it's here for a reason.

daattali:
I trust you on the gemfile thing - again, I don't really deal with ruby/gems, so I have to take your word for it.

I did see a PR for the round avatar, but I didn't see a separate PR for the config default values, those changes still seem to be in this PR

skalee:
> Thanks for letting me know about the 4.1 Jekyll roadmap - it does look like soon enough the `_data` issue won't be an issue anymore.

FYI `github-pages` gem is still using Jekyll 3.x, and it doesn't look like they're going to upgrade soon.

daattali:
That's ok - at some point in the future it will be upgraded, and until then it's fairly simple to have one sentence in the config.yml in the social icons section saying that if you're using remote_theme, you just need to copy this file

gpotter2:
> but I didn't see a separate PR for the config default values, those changes still seem to be in this PR

The commits are smashed together. It makes sense IMHO for it to be included here. (also it's a pain to split commits)

gpotter2:
Also it seems the data feature was removed from the 4.1 roadmap 6 days ago.. so I won't expect anything from there

daattali:
It is indeed a pain to split a commit, but it's worthwhile. A commit or a PR should be about one feature, so that if there's any issue, it's clear where and when exactly it happened, and it's easier to test a smaller more specific commit that does only one thing.

Adding default values must be its own PR so that it can be properly tested on its own. You don't have to do that PR if you don't want to, no obligation of course! But it can't go in here

Regarding the roadmap - I also don't see it _data folder there anymore. I'm going to suggest a different solution: get rid of the `_data/` folder altogether. 

- The ui-text data that staticman uses isn't **that** important (it's a neat feature but not a must, especially since staticman itself is not commonly used). 
- The SocialNetworks can be removed. Until a year ago, all the social network links were rendered in the footer without the help of a YAML file, and they YAML file came about as a way to reduce the repetitive code of creating the HTML for each social network - it was done [in this commit](https://github.com/daattali/beautiful-jekyll/commit/c2be2a963cc6610547539c00c63dda96f2569266#diff-1ccc2accd0fbe6ce53f777b86475a7b2). I do prefer the new method of using a for loop with a small HTML template, and I really hate copy-paste code, but I think in this case it's worthwhile to go back to the old copy-paste method of rendering the social networks, just so that the `_data/` folder can be eradicated because it's not fully supported. This should be its own PR.

gpotter2:
Commits was moved to a different PR.
This makes no sense because this PR still depends on it, but as you wish.

>The SocialNetworks can be removed. Until a year ago .......

Feel free to do so in a different PR. This one could still be merged without it

daattali:
I prefer the default values as a separate PR because it's a feature on its own, it's an independent issue (that happens to be a prereq here, yes). If there are any issues that are discovered with it down the line, it'll be easier to debug a commit that has only that rather than a commit that has other unrelated changes. Just makes for a cleaner git history, and I really try to keep clean histories for maintainability.

I'll deal with the data folders

daattali:
I refactored _data/SocialNetworks.yml so that it's no longer a data file. So now the only file that won't be copied over to a theme is the `ui-text` for staticman, which is ok with me since it's not a commonly used feature, we could just add a few words in the staticman section of the config file "if using remote_theme, you will need to copy the `_data/ui-text.yml` to your repository"

Otherwise we can move forward with this PR

daattali:
There should also be a mention that the `paginate` field is required in the config file, because without that your index page doesn't work. Any other config variables are useful but not required.

gpotter2:
Rebased & fixed. Also added the mentions you asked for

daattali:
Gotta resolve the main.css conflict that you created in your PR from today :)

README.md
```diff
@@ -171,6 +171,19 @@ googlefonts | List of Google fonts to include in the page (eg. `["Monoton", "Lob
 gh-repo   | If you want to show GitHub buttons at the top of a post, this sets the GitHub repo name (eg. `daattali/beautiful-jekyll`). You must also use the `gh-badge` parameter to specify what buttons to show.
 gh-badge  | Select which GitHub buttons to display, available options are: [star, watch, fork, follow]. You must also use the `gh-repo` parameter to specify the GitHub repo.
 
+### Use remote-theme (advanced)
+
+It is possible to use the `remote-theme` feature of GitHub pages. To do so, instead of forking this project:
+
+- Follow all the steps in https://guides.github.com/features/pages/ ***EXCEPT*** the ones where you chose the theme.
+- add `remote_theme: deanattali/beautiful-jekyll` to your `_config.yml`
```
daattali, line 179: daattali not deanattali

README.md
```diff
@@ -171,6 +171,19 @@ googlefonts | List of Google fonts to include in the page (eg. `["Monoton", "Lob
 gh-repo   | If you want to show GitHub buttons at the top of a post, this sets the GitHub repo name (eg. `daattali/beautiful-jekyll`). You must also use the `gh-badge` parameter to specify what buttons to show.
 gh-badge  | Select which GitHub buttons to display, available options are: [star, watch, fork, follow]. You must also use the `gh-repo` parameter to specify the GitHub repo.
 
+### Use remote-theme (advanced)
+
+It is possible to use the `remote-theme` feature of GitHub pages. To do so, instead of forking this project:
```
daattali, line 176: remote_theme not remote-theme

gpotter2:
All done.

daattali:
Is there any way for me to test this? Do you have a github repo set up with this that I can use as a remote?

gpotter2:
Yes, I'm using the master branch of https://github.com/gpotter2/beautiful-jekyll to host this PR.
`remote_theme: gpotter2/beautiful-jekyll` works.

daattali:
How are you using this PR as a theme? What value do I use as `remote_theme` in the config file to point to this pull request? I couldn't figure it out easily

gpotter2:
This PR is the master branch of https://github.com/gpotter2/beautiful-jekyll.

daattali:
Is there a way to automatically copy over the index.md file? Because right now if I create a repo and use remote_theme and add the config file, I'll get a 404. An index file is required, unless there's something I'm missing?

gpotter2:
A theme isn't really supposed to add an index page :/ no its not possible, at least as far as i know.

daattali:
You're right. I've tested it with GitHub Pages, looks good. How do I test it with local jekyll installations? Or does this only work with GH Pages and we need to make it a gem in order to use locally?

gpotter2:
If you're testing a github pages project locally, follow the instructions on https://github.com/github/pages-gem/blob/master/README.md (you'll need to create a Gemfile).

If your testing a project outside of github pages, you can manually use https://github.com/benbalter/jekyll-remote-theme

skalee:
I confirm that at least in case of my site all works as of 56c4cea7f36386a3e1c63eed5e84123a64de7a0a.

daattali:
I didn't like that if I run `jekyll new` and then add this theme, the homepage doesn't work at all, because the default jekyll assumes a `home` layout in the index page.

So I created a `home` layout and use it now in the index page, so that if someone uses `jekyll new` and wants to use this theme, they'll be able to see something immediately. I'll continue testing and hopefully merge tomorrow.

@gpotter2 @skalee do you know what's required to make this into a proper Ruby gem?

gpotter2:
You probably just need to update the `Gemfile` and add a gemspec.

Just another thought: remote theme has no versioning, so please note that once it goes live, you will be in charge to make sure any update you do is always backward compatible. If you want to make breaking changes without affecting users (change how the config values work...), now is the time.

daattali:
Good point, thanks for making that clear. I'm glad I spent a lot of time this week overlooking the entire config file.

gpotter2:
Great thanks for the merge. (I also noticed I forgot to add a credit :smile:)

daattali:
I added both of you to the changelog

---
Dean Attali
President & CEO
AttaliTech Ltd
http://AttaliTech.com <http://attalitech.com>

skalee:
> Just another thought: **remote theme has no versioning**, so please note that once it goes live, you will be in charge to make sure any update you do is always backward compatible. If you want to make breaking changes without affecting users (e.g. change how the config values work, for instance you just changed how "author" was stored), now is the time.

It does support versioning, see https://github.com/benbalter/jekyll-remote-theme#declaring-your-theme. For instance, to freeze on what's currently on master, following could be specified: `daattali/beautiful-jekyll@790bbf59a248596317a47353c45b4823d7fcf471` And with Git tags it can be much more user-friendly.

gpotter2:
Nice catch.
Git tags are a good idea. For simplicity you probably want to start the versioning where the gems left off.

daattali:
Thanks, I'll make a tag indeed.

daattali:
@gpotter2 I just noticed that there are two instances where you didn't migrate the js/css into assets:

https://github.com/daattali/beautiful-jekyll/blob/master/_includes/staticman-comments.html#L77-L81

and

https://github.com/daattali/beautiful-jekyll/blob/6552c5774efec8abbbbf0977a13a619998c77fef/_includes/head.html#L130

Is this correct, was this done on purpose for any reason?

gpotter2:
No i missed those.

daattali:
Everything has been addressed, and it works great. @skalee in the documentation I wrote that they should use `beautiful-jekyll-theme@2.3.0` and that every so often they can check for new versions

gpotter2:
Looking good, great job 🎉

PR #591, opened 2020-03-25
add new comments - gitalk support

colynn:
## Background
For some reason, we can't use `disqus`/`staticman` as a comment tool very well, but all we have github account, and I find Gittalk[Gitalk is a modern comment component based on Github Issue and Preact](https://gitalk.github.io),  so I want to share it with us.

## Changes

1.  add `gittalk-comment.html` into `_includes`
2. change post.html/ page.html link to gittalk-comment.html
3. add Gittalk parameters into `_config.yml`  and `README.md` docs

gittalk

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,8 +120,8 @@ To use Staticman, you first need to invite `staticmanlab` as a collaborator to y
 
 Optional: You may want to configure a webhook to prevent old inactive branches (representing approved comments) from stacking up.  You can refer to [Staticman's documentation](https://staticman.net/docs/webhooks) for details.  Make sure to input the **Payload URL** according to your chosen `endpoint`.  For example, the default `endpoint` is `https://staticman3.herokuapp.com/v3/entry/github/`, so the corresponding **Payload URL** should be `https://staticman3.herokuapp.com/v1/webhook`.
 
-#### Gittalk comments 
-To use gittalk, you first need to register a gitHub application by going to register github application (https://github.com/settings/applications/new), **Note** You must specify the website domain url in the Authorization callback URL field.  Lastly, fill in the github application's `Client ID`、`Client Secret` and `gittalk` parameters in the Gittalk section of `_config.yml`.
+#### Gitalk comments 
+To use gitalk, you first need to register a gitHub application by going to register github application (https://github.com/settings/applications/new), **Note** You must specify the website domain url in the Authorization callback URL field.  Lastly, fill in the github application's `Client ID`、`Client Secret` and `gitalk` parameters in the Gittalk section of `_config.yml`.
 
 ### Adding Google Analytics to track page views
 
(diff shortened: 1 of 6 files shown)


colynn:
@daattali  Anything else to add， Could you accept the pull request?

daattali:
@colynn thanks for the initiative!

Out of curiousity, what do you mean "For some reason, we can't use disqus/staticman"? Is it blocked in china, you're actually unable to use these tools?

I generally try to remain very conservative and not add too many third party tools, because each new tool adds an extra layer of complexity and maintenance. I would be happy to add this tool if you think it's useful to a lot of people, simple, and is the best of its alternatives (I don't want to add multiple similar tools because I want this theme to remain simple to support people who are absolute beginners). Have you considered alternatives like https://github.com/imsun/gitment and https://utteranc.es/ ? 

Thanks again

colynn:
[gitalk](https://github.com/gitalk/gitalk) is similar with [gitment](https://github.com/imsun/gitment),  so now I should close this pull request? @daattali

daattali:
No, I'm asking for your opinion. What do you mean that you can't use disqus
and staticman? And if you can't use those - have you looked at similar
systems and determined that gitalk is the best one? I don't know, I haven't
looked into them - I'm asking because this theme affects thousands of
people so before any additional feature I want make sure it gets proper
consideration :)

colynn:
- Firstly, Gitalk is better than Gitment, and Gitment looks like it hasn't been maintained for a long time,

- Gitalk easy do `Quote reply` but utteranc.es not support this.

- When new users need to leave comments should be at the top instead of the bottom, base on this, Gitalk is better than utteranc.es. But Gitalk did not pick your reaction, but  utteranc.es support this.

-  For now, I will choose Gitalk.  But I think utteranc.es has more potential, and I'd like to pause this pull request.

daattali:
I do like the idea of this tool. I think the documentation for it is not 100% clear, I didn't understand what the `repository` parameter - do I need to create a new public github repository with Issues enabled that will store the comments, and provide the name of the repository? And what do the `admins` mean exactly, why is that needed for the comments system? Does that mean people who can go to the github repository and delete comments?

colynn:
> I do like the idea of this tool. I think the documentation for it is not 100% clear, I didn't understand what the `repository` parameter - do I need to create a new public github repository with Issues enabled that will store the comments, and provide the name of the repository? 

Yeah,  you can base on exist public github repository or create a new one that will store the comments, then create github outh app follow github [OAuth flow](https://developer.github.com/v3/oauth/#web-application-flow).

> And what do the `admins` mean exactly, why is that needed for the comments system? Does that mean people who can go to the github repository and delete comments?

I think `admins` is additional functionality, only these guys can initialize github issues issues(meaning create an github issue for this page), but utteranc.es do initialize by [utteranc.bot]https://github.com/utterances-bot).

About this part, Gitalk is not a good practice.

daattali:
I appreciate your time @colynn 

I would be happy to accept a PR for utterance, it seems more stable and robust

Issue #593, opened 2020-03-29
How to display the time of reading on the blog


memoriesus:
Thank you for submitting an issue!

Please only submit bug reports or feature suggestions. Please do not submit support requests and general help questions in this forum.

daattali:
Can you add details of what you mean?

memoriesus:
I mean it can display the time to tell the reader the time you can read it over.

daattali:
That's not a supported feature

rubyreads:
@memoriesus 
There is code that you can add, if you're comfortable doing so. Here is an [example](https://gist.github.com/rubyreads/b1156a4bea4e1d5ed823b87260d410f5). Since it is not a supported feature, you'll have to:
- decide where to place it
- make sure it doesn't conflict with the Beautiful Jekyll theme  

I hope this helps.

daattali:
Thanks @rubyreads ! I'd be happy to add this feature if you want to submit a pull request

memoriesus:
But based on it it's not available to show me the real-time.

daattali:
What do you mean real time?

memoriesus:
I mean  now the code cannot make it happen, perhaps it has a bit bug.

daattali:
I don't understand what you mean. Currently this is not supported. @rubyreads is in the process of adding this feature.

daattali:
Fixed with #622

PR #594, opened 2020-03-29
Add readtime

rubyreads:
Feature suggestion - "How to display the time of reading on the blog #593"

This file makes it possible to use {% include readtime.html %} wherever an estimated read time is desired (such as next to post date).

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -151,7 +151,7 @@ title-separator: "-"
 date_format: "%B %-d, %Y"
 
 # Read Time Option
-readtime: true # Change to true to include estimated reading time on posts
+read-time: true # Change to true to include estimated reading time on posts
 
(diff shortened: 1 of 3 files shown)


daattali:
Thank you @rubyreads 

In order to add this feature, we would want to really integrate it , rather than leave it in the source code as something undocumented that will get hidden. Could you add this in an appropriate location in the post header, along with some appropriate CSS, and add a config param to turn this on/off (default to off)? The goal is that anyone forking this theme will be able to intuitively and easily find this and enable it

rubyreads:
@daattali -- this is my first pull request on a public repo... I'll do my best. Thank you 🙂

memoriesus:
But maybe it is not available

rubyreads:
I created a new pull request #595 (which has all the necessary commits).

PR #595, opened 2020-03-30
Improved pull request for "Add readtime #594"

rubyreads:
@daattali -- it took me a while to figure this out; I ended up starting over with a new fork to make things clean. I originally had trouble because of nested if statements and improperly adding the config param to work elsewhere as an invalid  `site.readtime` instead of making it able to work with `page.readtime` (creating a failure to build). However, I caught these mistakes and the code ended up working on each post.

**Note**: Readtime will only appear on each post if enabled in _config.yml; it will not appear on index.html (that is, the list of posts). People may want the readtime to appear next to the post date on that list though. I attempted to add the code to that page but I was unable to make it work; it is beyond my limited experience.

PR #597, opened 2020-04-01
make the header display the read time of posts

tirthasheshpatel:
tried to make the header display the read time of posts.


tirthasheshpatel:
never mind. just saw #595 solves it

PR #608, opened 2020-04-13
Allow configuring lang attribute on <html> tag

skalee:
Add a page/site configuration variable named `language`, which can be set either in `_config.yml` or in a front matter in order to customize value of `lang` attribute on `<html>` element.  Previously, it was hardcoded to `en`.


daattali:
Looks good. Could you move the documentation for it lower in the table, because I don't think it's more important than a lot of other more commonly used variables. And could you also add an entry in the CHANGELOG file with an attribution to yourself

daattali:
@skalee I'd be happy to merge with these two small changes

skalee:
@daattali 

- option description has been moved down so that it's now right after description of `use-site-title`
- changelog entry has been added

I am not sure if that's what you requested, please be more specific if not.

daattali:
Perfect, thanks

PR #612, opened 2020-04-18
Allow to disable the round avatar

gpotter2:
- adds ability to turn off the round avatar

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ This file lists all non-trivial changes to Beautiful Jekyll.
 
 I often make small changes to documentation, to the demo site, or to the general look-and-feel. These changes will not be listed here. Any other minor changes will also not be listed here.
 
+**2020-04-18** Add ability to disable round logo.
+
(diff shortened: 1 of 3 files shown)


gpotter2:
I didn't realize there is no CI on this project

daattali:
Good point, a CI (Travis? Github actions?) could be a good addition

daattali:
Can you just add an entry to the CHANGELOG file please? Then this will be merged

daattali:
Have you tested this code?

I just tried it and I'm getting 3 similar errors

> Liquid Warning: Liquid syntax error (line 174): Expected dotdot but found pipe in "(site.round-avatar | default: true)" in css/main.css

Does it work for you without errors?

daattali:
I need to revert this PR, it doesn't work correctly. If I remove the avatar-round field from the config file, it isn't round. Did you test this before submitting?

gpotter2:
Sorry for that. I couldn't publish the 2 PRs at the same time. That's why I didn't want to split them.

PR #621, opened 2020-04-20
Add default values

gpotter2:
see https://github.com/daattali/beautiful-jekyll/pull/590

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -25,7 +25,7 @@
           {%- endfor -%}
         </ul>
       <p class="copyright text-muted">
-      {% if site.author %}
+      {% if site.author.name %}
       {{ site.author.name }}
       &nbsp;&bull;&nbsp;
       {% endif %}
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -25,8 +25,10 @@
           {%- endfor -%}
         </ul>
       <p class="copyright text-muted">
+      {% if site.author %}
       {{ site.author.name }}
       &nbsp;&bull;&nbsp;
+      {% endif %}
       {{ site.time | date: '%Y' }}
 
       {% if site.url-pretty %}
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -3,7 +3,7 @@
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
   <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, viewport-fit=cover">
 
-  <title>{% if page.use-site-title %}{{ site.title }} {{ site.title-separator }} {{ site.description }}{% elsif page.title %}{{ page.title }}{% else %}{{ site.title }}{% endif %}</title>
+  <title>{% if page.use-site-title %}{{ site.title }} {{ site.title-separator | default: '-' }} {{ site.description }}{% elsif page.title %}{{ page.title }}{% else %}{{ site.title }}{% endif %}</title>
 
   <meta name="author" content="{{ site.author.name }}" />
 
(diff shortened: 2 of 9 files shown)


daattali:
Thanks! There are still instances of config variables unaccounted for in the code, for example I know for certain that date_format is also in `_includes/header.html`. The codebase needs to be fully scanned for all config variables

gpotter2:
You're right. I'll grep to make sure i don't miss anything

gpotter2:
I've grepped the project. This should be all of them.
The few i left do not need a default value

_includes/footer.html
```diff
@@ -25,8 +25,10 @@
           {%- endfor -%}
         </ul>
       <p class="copyright text-muted">
+      {% if site.author %}
```
daattali, line 28: This should be `if site.author.name`

It's possible to define `author` but not a subfield `name` , in which case a bullet point will show up

_includes/head.html
```diff
@@ -3,15 +3,15 @@
...
+  <title>{% if page.use-site-title %}{{ site.title }} {{ site.title-separator | default: '-' }} {{ site.description }}{% elsif page.title %}{{ page.title }}{% else %}{{ site.title }}{% endif %}</title>
 
   <meta name="author" content="{{ site.author.name }}" />
 
   {% if page.subtitle %}
   <meta name="description" content="{{ page.subtitle }}">
   {% endif %}
 
-  <link rel="alternate" type="application/rss+xml" title="{{ site.title }} {{ site.title-separator }} {{ site.description }}" href="{{ '/feed.xml' | absolute_url }}" />
+  <link rel="alternate" type="application/rss+xml" title="{{ site.title | default: 'Untitled' }} {{ site.title-separator }} {{ site.description }}" href="{{ '/feed.xml' | absolute_url }}" />
```
daattali, line 14: `title-separator` needs a default here too

_includes/head.html
```diff
@@ -3,15 +3,15 @@
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
   <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, viewport-fit=cover">
 
-  <title>{% if page.use-site-title %}{{ site.title }} {{ site.title-separator }} {{ site.description }}{% elsif page.title %}{{ page.title }}{% else %}{{ site.title }}{% endif %}</title>
+  <title>{% if page.use-site-title %}{{ site.title }} {{ site.title-separator | default: '-' }} {{ site.description }}{% elsif page.title %}{{ page.title }}{% else %}{{ site.title }}{% endif %}</title>
 
   <meta name="author" content="{{ site.author.name }}" />
```
daattali, line 8: this meta tag shouldn't be included if author name isn't available

_includes/header.html
```diff
@@ -1,5 +1,7 @@
 <!-- TODO this file has become a mess, refactor it -->
 
+{% assign date_format = site.date_format | default: "%B %-d, %Y" %}
```
daattali, line 3: `tags.html` also has a date_format that needs a default

_includes/head.html
gpotter2: I assumed if you use `use-site-title`, you had provided a title..

daattali:
Fair

_includes/head.html
gpotter2: I assumed using `use-site-title` meant there was title. Another check seems over the top

gpotter2:
Whoops I didn't mean to delete my comment.
Will fix those in a minute

daattali:
Message me when all done

_includes/footer.html
gpotter2: done

_includes/head.html
gpotter2: done

_includes/head.html
gpotter2: - if page-use-title is here it's enough
- done

_includes/header.html
gpotter2: done

gpotter2:
Should be done.

daattali:
I added one small commit, but otherwise looks good!

Issue #641, opened 2020-05-18
Build fails


rs9899:
I'm noticing there being an issue of exit code 15 in jekyll build. Is there any issue with that Jekyll yml file in the GitHub action workflow?

daattali:
Yes I did notice that, but I'm not sure why. I looked at the history of builds and can't tell why it started failing. If you're able to figure it out I would be happy to get some help.

The theme itself does work with no issues, it's only the github action that's broken.

rs9899:
Exit code 15 seems like out of resources
https://www.mersenneforum.org/showthread.php?t=23370

maybe its more of a resource constraint of the public github repo maybe. Not sure though. I will look more on this too.

daattali:
I haven't been able to figure it out. It would be very very helpful if you do -- thank you

PR #644, opened 2020-05-24
Fix CI

gpotter2:
- pin docker builder version

@daattali You'll just have to update this version each time you change the jekyll version in the gemspec.

fix https://github.com/daattali/beautiful-jekyll/issues/641


daattali:
Thanks! I would not have caught this. Did this break because github pages updated their jekyll gem?

Issue #646, opened 2020-05-28
[Proposal] Add Matrix and Mastodon social network links


Iolaum:
Hello, I have started using your template recently ([here](https://github.com/Iolaum/Iolaum.github.io), it's still barebones). I really liked it and I am going to spend more time with it and make a small website out of it.

I was exploring the `_config.yml` options, being pleasantly surprised by the many options that were available for [social network links](https://github.com/daattali/beautiful-jekyll/blob/54ca4023738b77d3b7c64f7aded7fc404f00fec7/_config.yml#L40), and noticed that there were no Mastodon and Matrix links. Given all the other ones that are there, I am wondering if you would be open to considering including them in the template.

daattali:
I haven't heard of these servies, so I had to look them up and they seem legitimate. Feel free to submit a PR

daattali:
Matrix doesn't seem to have a font-awesome icon

daattali:
@Iolaum feel free to submit a PR if you'd like to see these added, but I'm closing the issue

PR #651, opened 2020-06-07
Adding Mastodon Social Network Link

Iolaum:
PR follows discussion at #646 .

Making a PR with only a mastodon link. Looking forward to any comments. A tricky thing is that because mastodon is decentralised there are two things that need to be specified, the hosting instance URL and the username on said instance. I went around this by suggesting the mastodon entry in `_config.yml` be filled in such way that both can be presented as one variable.

P.S. I am not adding a matrix link in this PR because a matrix logo is not included in font-awesome currently. I have [one implementation](https://github.com/Iolaum/beautiful-jekyll/tree/dev_social_matrix) that [uses another non-related picture](https://github.com/Iolaum/beautiful-jekyll/commit/46b1e9f5c7548b0097cc16379f974fbe6930a419) and  a [second implementation](https://github.com/Iolaum/beautiful-jekyll/tree/dev_social_matrix_fa) that [uses fork-awesome to get a matrix logo](https://github.com/Iolaum/beautiful-jekyll/commit/483e40c562e24602134d4dc5ee93f06fdfe0e77e). I 'd be interested if any of those approaches could be considered for inclusion, but given their drawbacks I decided to not include them in the PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## v3.0.1 (Date: TBD)
 
 - Fixed issue where if a dropdown menu was the last item in the menu bar, it did not have a proper margin on the right
+- Add social network link: Mastodon
 
 ## v3.0.0 (Date: 2020-05-07)
 
--- a/_config.yml
+++ b/_config.yml
@@ -57,6 +57,7 @@ social-network-links:
 #  yelp: yourname
 #  telegram: yourname
 #  calendly: yourname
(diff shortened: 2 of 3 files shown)


daattali:
Looks good. Please update the CHANGELOG too

I would prefer not to complicate the framework by including non font awesome icons

Iolaum:
Makes sense, looks like the Matrix link can wait.

Added changelog entry.

PR #659, opened 2020-06-13
Web monetization

artis3n:
This adds support for the new [Web Monetization proposed standard](https://webmonetization.org/).

A very useful article to get started is <https://dev.to/hacksultan/web-monetization-like-i-m-5-1418>, so that is linked from the config option.


daattali:
Thank you for the initiative, but I prefer not to add this to the theme

PR #662, opened 2020-06-18
Hugo implementation

SamEdwardes:
Changing from Jekly to Hugo theme


SamEdwardes:
Oops, did not mean to send the PR this way!

PR #670, opened 2020-07-05
Add orcid and google scholar

hsleese:
Added support for google scholar and ORCID social networks.


hsleese:
See https://materialsforhealthlab.org for how it looks.

daattali:
Thanks!

PR #690, opened 2020-08-07
Adds page configuration for description

lexicalunit:
Sometimes a page subtitle isn't substantive enough to be a good value for a meta description. An optimal length for meta descriptions is usually around 150 characters, which might be too long for a pithy page subtile. In this case I added a `site.description` override that will be used only if provided. Otherwise we fallback to using the subtitle.


lexicalunit:
Getting a weird dependency error in the tests. Any ideas @daattali?

daattali:
Regarding the build error, no idea. @gpotter2 any clue?

daattali:
If I don't set a description on a page, but I have a description defined in the config.yml, then it will always override the subtitle, right?

lexicalunit:
As far as I know: No. It's my understanding that `page.description` would only ever come from page front matter. So under no circumstances could anything in `_config.yml` override a page subtitle.

daattali:
I just tested this, and you're correct.  `page` params do not inherit from `site`. Thanks

gpotter2:
He was faster than me and spotted the missing dependency :-)

daattali:
Thanks for taking a look anyway!

PR #691, opened 2020-08-07
Adds support for meta keywords

lexicalunit:
Adds support for meta keywords. It will use the page keywords if provided, otherwise fallback to site level.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Unreleased version
 
+- Adds support for site level keywords page meta header. (#691)
 - Adds support for custom meta-theme site color. (#692)
 - Add support for `site-css` config parameter to provide site-level CSS files. (#695)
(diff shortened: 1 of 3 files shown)


daattali:
Good idea! Right now there is no keywords support at all. Do you strongly feel the need for both side-wide and page-specific keywords?

lexicalunit:
My use case is only for site-wide but I thought I'd toss in page-specific support just in case. I don't feel strongly about it.

daattali:
Let's keep it simple at site-wide only. I'm trying to active fight getitng a huge laundry list of options.

daattali:
Clarification: I'm sure there will be *some* people who want even more customization per-page. But having keywords globally should be what the vast majority of users use, and I prefer simplicity over catching 100% of usecases

lexicalunit:
Right on. Updated.

daattali:
The documentation and the README need to be updated accordingly

lexicalunit:
Oops! My bad.

PR #692, opened 2020-08-07
Adds support for meta theme-color

lexicalunit:
Adds support for [meta theme-color](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name/theme-color).


daattali:
In an effort to make the config file readable to the average user, could you change the name of the param and the documentation to something that a lay person would understand?

lexicalunit:
Can do! PR updated.

PR #693, opened 2020-08-07
Adds configuration for site font family

lexicalunit:
I noticed there wasn't any way to configure the site font-family so I added it.


daattali:
I appreciate the initiative, but this is something I prefer not to add as a parameter. It's too specific and I suepect the usage of it will not be high, and it really blurs the line between css and config file

PR #694, opened 2020-08-07
Adds configuration for text-color

lexicalunit:
Mirroring the existing `site.page-col` configuration, I added one for text color called `site.page-text-col`.


daattali:
I would change the name to simply `text-col`. And there are a few other uses of #404040 in the file, does the page look fine when those two are kept as 404040 or should they also match?

daattali:
Could you add all these changes that you're adding to the CHANGELOG?

lexicalunit:
@daattali I was wondering if I should be adding these changes to `CHANGELOG` or not. I wasn't sure about the format. Should I add an "[Unreleased]" section?

daattali:
I just updated the changelog to givey ou a palce to add these. One bullet point per PR

lexicalunit:
Updated!

daattali:
Did you notice my [first comment](https://github.com/daattali/beautiful-jekyll/pull/694#issuecomment-670802292)?

lexicalunit:
> Did you notice my [first comment](https://github.com/daattali/beautiful-jekyll/pull/694#issuecomment-670802292)?

Nope! Missed it. You're right that those other locations also need to be updated to use `text-col`. I'll update in a sec.

PR #695, opened 2020-08-07
Adds configuration for additional site css

lexicalunit:
Having `page.css` is nice, but there's no way to add css for the entire site. You could just modify `main.css` with the changes you want, but that makes upgrading to the latest from upstream somewhat difficult as you have to deal with potentially complicated merge conflicts. Or you could add the additional css to each and every page, which is annoyingly tedious and human-error prone.

This PR adds a site level configuration for `site-css` that allows you to list additional css files that should be included for every page on the site.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -101,6 +101,10 @@ footer-link-col: "#404040"
 #footer-img: "/assets/img/bgimage.png"
 #page-img: "/assets/img/bgimage.png"
 
+# You can also include any number of additional CSS assets in every page on your site
+#site-css:
+#  - "/assets/css/my-style.css"
+
 # --- Web Statistics Section --- #
 
 # Fill in your Google Analytics gtag.js ID to track your website using gtag
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -42,6 +42,12 @@
     {% endfor %}
   {% endif %}
 
+  {% if site.site-css %}
+    {% for css in site.site-css %}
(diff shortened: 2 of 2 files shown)


daattali:
Thanks for the suggestion! It's a good idea, but if it's a site-wide configuration option, then it should really be in the config.yml in my opinion. If it's in the YAML front mattter, then every page has to define it separately. Am I correct or am I misunderstanding your intention?

lexicalunit:
I think we agree! That's why I used `site.site-css` as opposed to `page.site-css`. So that the configuration takes place in `_config.yml` instead of needing to be in the page front mater.

daattali:
In that case, the setting should be removed from the README (the readme section only describes settings that are used in the YAML since there is nowhere else they are documented), and added to the config file

lexicalunit:
Ah, noted! PR updated.

daattali:
I still don't see it in the config.yml file?

lexicalunit:
Oh, like add a placeholder there with documentation? Of course. I'll make that change for this PR and the other ones.

daattali:
Yes, exactly. Look at the config file, you'll see lots of options that are disabled either by being commented out or by having a blank value. And the documentation for them is just comments beside each option.

daattali:
Great, thank you!

daattali:
I made a small change, I moved the site wide css files to be included before any page-specific files, so that css rule cascading works more intuitively

PR #697, opened 2020-08-08
Kramdown parser gfm dep

lexicalunit:
I think this will resolve the build errors we're seeing on the latest PRs. I think one of the existing runtime dependencies must have had `kramdown-parser-gfm` as a dependency, but doesn't any longer? So adding it to our dependency list resolves the issue.


daattali:
Nice detective work :)

PR #699, opened 2020-08-08
Adds og:site_name support

lexicalunit:
Adds meta tags for `og:site_name` which will just use the existing `site.title` configuration. Slightly different than `og:title` which can be overridden on a per-page basis. Whereas `og:site_name` will _always_ just be whatever `site.title` has been configured to be.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -63,6 +63,10 @@
   <meta property="fb:app_id" content="{{ site.fb_app_id }}">
   {% endif %}
 
+  {% if site.title %}
+  <meta property="og:site_name" content="{{ site.title }}">
+  {% endif %}
+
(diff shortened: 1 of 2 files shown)


_includes/head.html
```diff
@@ -71,7 +75,7 @@
   <meta property="og:title" content="{{ site.title }}">
   {% endif %}
 
-   {% if page.meta-description %}
```
lexicalunit, line 74: Fixes a slight indentation misalignment here.

_includes/head.html
```diff
@@ -71,7 +75,7 @@
...
+  {% if page.meta-description %}
```
lexicalunit, line 78: Fixes a slight indentation misalignment here.

PR #700, opened 2020-08-08
Cleans up trailing whitespace

lexicalunit:
Just an easy little PR to clean up some unhelpful trailing whitespaces in various files. This stuff can be particularly problematic when trying to maintain clean diffs in the face of most code editors these days which helpfully strip whitespace on file save.


daattali:
This is useful, thanks! This is my only project where I often edit files directly on github or notepad instead of a proper IDE, so this happens... 😳

PR #702, opened 2020-08-12
Automatically set navbar light/dark based on color

lexicalunit:
### Description

Pretty simple little bit of javascript to intelligently set the navbar class to either `navbar-light` or `navbar-dark` based on the color of the navbar that the user has configured. This will make it so that navbar elements like the hamburger menu in the mobile site view are easier to see on dark background colors.

### Examples

| Before | After | Result |
| --- | --- | --- |
| before-spellbot | after-spellbot | actually visible now | 
| before-darkblue | after-darkblue | better visibility |
| before-ivory | after-ivory | no difference |
| before-maroon | after-maroon | better visibility |
| before-pink | after-pink | no difference |

As you can see, for any light colored navbars there's is no difference at all because we will pick `navbar-light` which was the previous default. But for the darker colors the button changes from either completely invisible to actually visible or from hard-to-see to easier-to-see.

### Benefits:

- Users who want to use dark navbar colors are now automatically handled correctly.
- Users don't have to learn about [bootstrap navbar color schemes](https://getbootstrap.com/docs/4.0/components/navbar/#color-schemes) to get dark colored navbars to work.
- Users don't have to modify `_includes/nav.html` to get dark colored navbars to work.
- Users don't have to consider modifying their site css/config files to get their navbar to look good.

Overall this helps non-technical or even just somewhat-technical users of `beautiful-jekyll` to easily have their website do the right thing when they make a trivial configuration change to the navbar color, such as just changing it to a darker color.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased version
 
-- Add YAML option `head-extra` which is similar to `footer-extra` but is used to include custom HTML code in a page's `<head>` tag 
+- Add YAML option `head-extra` which is similar to `footer-extra` but is used to include custom HTML code in a page's `<head>` tag
+- Add automatic navbar color detection (#702)
 
 ## v4.1.0 2020-08-08
 
--- a/assets/js/main.js
+++ b/assets/js/main.js
@@ -7,14 +7,14 @@ var BeautifulJekyllJS = {
 
   init : function() {
     // Set the navbar-dark/light class based on its background color
-    const rgb = $('.navbar').css("background-color").replace(/[^\d,]/g,'').split(",")
+    const rgb = $('.navbar').css("background-color").replace(/[^\d,]/g,'').split(",");
     const brightness = Math.round(( // http://www.w3.org/TR/AERT#color-contrast
       parseInt(rgb[0]) * 299 +
       parseInt(rgb[1]) * 587 +
-      parseInt(rgb[2]) * 114) / 1000);
+      parseInt(rgb[2]) * 114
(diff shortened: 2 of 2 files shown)


daattali:
Thanks for bringing this to my attention, I didn't notice that changing the colours doesn't affect the hamburger menu. Could we simply change the colour of that menu to match the navbar-text-colour?

That's a beautiful table you added to the issue BTW!

lexicalunit:
I played with it a bunch to try and figure out how to fix it until I finally stumbled upon that bootstrap documentation page and found the `navbar-light` class. There might be another way to fix it but I didn't find one. I am not the best web dev tho! So it might be worth investigating a bit if you wanna 👍

daattali:
Does your solution work if a colour is given not as hex? Using `red` or `rbg(255,0,0)` instead of `#ff0000`?

lexicalunit:
Good question! It does. I was worried about that too, but thankfully Jquery’s implementation of `.css(“background-color”)` will always return a string which contains either “rgb(...)” or “rgba(...)” so I made sure the regex pattern would match either one of those possibilities.

daattali:
Oh that's great. I would prefer to leave the HTML with the light version and then javascript can remove and re-apply. So that if for any reason a javascript error happens and that line of javascript never gets to run, there will still be something applied (without it, the menu doesn't show up)

lexicalunit:
Good idea, lemme make that change right now.

daattali:
Awesome. Can you add to CHANGELOG? :)

lexicalunit:
Done and done!

PR #723, opened 2020-10-01
Update 404.html

PranjalTripathi1711:
Closing not required on break tag.


daattali:
Thanks. There are other instances in the codebase that close a br tag, do you mind fixing them all with a simple grep?

PranjalTripathi1711:
> Thanks. There are other instances in the codebase that close a br tag, do you mind fixing them all with a simple grep?

Not a problem I will have a look at that.

daattali:
Do you want to return to this?

Issue #744, opened 2020-11-24
Quotes in a post title cut the title in the social preview


f-squirrel:
I've noticed that if a post has quotes in its title then when the post is shared on Facebook or Twitter, only the part before the quotes is shown.
Example:
```
title: The "moving" truth behind std::optional
```
=> The preview shows everything before the quote:
```
The
<excerpt>
<share-img>
```
When I remove the quotes:
```
title: The moving truth behind std::optional
```
=> The preview shows the full title:
```
The moving truth behind std::optional
<excerpt>
<share-img>
```
I've tried to put `\"` but it did not help.
 
I am not sure if it is a bug or misconfiguration, I'd be happy if anybody has any leads to offer.

P.S. I am really grateful for this awesome theme and the whole work that Dean and the contributors are doing!

daattali:
This seems like an escaping issue, I'm probably not escaping the title somewhere where it should be escaped. I'd be happy for a PR!

f-squirrel:
@daattali , I've created a PR(https://github.com/daattali/beautiful-jekyll/pull/745).
I am pretty new to HTML/Ruby, but it worked for me.

PR #745, opened 2020-11-24
fix(#744): Escape special characters in title

f-squirrel:
Fixes https://github.com/daattali/beautiful-jekyll/issues/744

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Fixed bug where using an image as a navbar title did not render in GitHub Project pages that did not have a custom domain
 - Fixed issue where image thumbnails on the feed page were always forced into a square rather than maintaining a proper image aspect ratio
 - Added support for Patreon in the social network links in the footer
+- Fixed bug where special characters in the title led to broken share tags(#744)
 
 ## v5.0.0 (2020-09-15)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 - Fixed bug where using an image as a navbar title did not render in GitHub Project pages that did not have a custom domain
 - Fixed issue where image thumbnails on the feed page were always forced into a square rather than maintaining a proper image aspect ratio
 - Added support for Patreon in the social network links in the footer
-- Fixed bug where special characters in the title led to broken share tags(#744)
+- Fixed bug where special characters in the title led to broken share tags (#744)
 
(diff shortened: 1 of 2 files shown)


daattali:
Thanks!

Are you sure we want that in _includes/header.html as well? Did you have an example of what happens currently on the page and how it's fixed with your PR?

f-squirrel:
I am not, I've only made sure that special characters used in `title` are escaped whenever is possible.
Share card before the fix:
Screen Shot 2020-11-25 at 14 31 01
After the fix:
Screen Shot 2020-11-25 at 14 34 38

From an HTML perspective, all `"` used in title are replaced with `&quote;`.

The link to the fixed post:
https://ddanilov.me/the-state-of-std-optional-after-move/

daattali:
Yes I agree this is needed in the meta tags. But the other parts of the code that you changed, unless they're tested and you know that they're needed, we should remove them.

f-squirrel:
@daattali , The link I've posted contains a page with the fix implemented. Hence it is tested.
If the title contains other special characters like `<`, `>` or `&` and the header.html file is not changed, then the title is shown only partially.
Example:
`title: The <moving" truth behind std::optional`
Is rendered as following:
Screen Shot 2020-11-26 at 14 55 03
I think it might be safer to change the `_includes/header.html` as well.
WDYT?

daattali:
Are you having issues with anything besides double quotes?

Because I just spent some time reading on this, and it seems that for share cards, double quotes should be avoided, which means this isn't a beautifuljekyll issue. If you google for this, you'll find that it's generally a bad practice to include quotes in the title/description, and if you do have to use quotes then you should use single rather than double quotes.

daattali:
Regardless, I do agree with stripping html tags and escaping in the title/description meta tags. But either way, a double quotation mark in the title is not generally a good idea.

In the title of the page, currently HTML tags do work if you put quotes around. For example `title: "one <i>two</i> three" will show the word two in italics. I don't know if this behaviour is good or not - should titles be disallowed from using html tags? What do you think?

f-squirrel:
@daattali , agreed, HTML tags are a legitimate part of the title.
I've updated the PR(removed the changes in the `_includes/header.html` file.

Thank you for the detailed expalnation.

daattali:
Could you add an entry in the CHANGELOG please

f-squirrel:
Done.

Issue #749, opened 2020-12-04
Site avatar image is too low on Safari


lexicalunit:
I didn't dig into what's going on here, but it seems to work fine on Firefox and Chrome.

**large screen**
<img width="914" alt="Screen Shot 2020-12-03 at 10 42 31 PM" src="https://user-images.githubusercontent.com/1903876/101130833-0f7cef80-35b9-11eb-8e06-df7e1f887f43.png">

**small screen**
<img width="815" alt="Screen Shot 2020-12-03 at 10 42 44 PM" src="https://user-images.githubusercontent.com/1903876/101130829-0e4bc280-35b9-11eb-975d-726c220dd4c8.png">

daattali:
Thanks for the report. I don't have any macs or iphones around me, I'll try to look into it when I'm with someone who has an Apple device

daattali:
To confirm: this happens for all beautifuljekyll websites, not just the demo one, right?

lexicalunit:
It's happening for my site and I checked https://beautifuljekyll.com/ just to see if it was happening there too just to confirm. I didn't check more sites than that though.

daattali:
That sounds like it's probably true for all of them :)

PR #750, opened 2020-12-04
Fix margin top alignment issue on Safari

lexicalunit:
Resolves https://github.com/daattali/beautiful-jekyll/issues/749

~~Changing the margin-top just a bit seems to fix the issue on Safari while leaving Firefox and Chome nearly unchanged.~~

Edit: Updated the PR to use the better suggestion from @daattali.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Unreleased version
 
+- Slightly reworked margins and position for avatar image to resolve an alignment issue on Safari.
 - Changed the width at which the navbar collapses to a higher threshold because most modern non-mobile browsers are >1000px
 - Fixed bug where navbar secondary level dropdown items didn't inherit te same colour as the primary navbar links
 - Fixed bug where the navbar "burger" collapsed button didn't always revert back to a light colour
--- a/assets/css/beautifuljekyll.css
+++ b/assets/css/beautifuljekyll.css
@@ -311,7 +311,7 @@ img {
   position: absolute;
   left: 50%;
   width: 3.125rem;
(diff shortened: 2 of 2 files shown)


daattali:
Thanks Amy

I tested this out but it does seem to cause a larger visual change than I'd be comfortable with. Is this something that only broke in the latest version, or has it been the case before too? I'll try to debug it a different way, I don't want to change the spacing that much for unaffected browsers. 

By the way, would you like your spellbotio website featured , as a thank you for your contributions?

lexicalunit:
Yeah I only noticed it after updating to version 5 E
earlier today. Previously it was looking ok on Safari.

And yah, a call out would be totally welcome!

daattali:
Could you please test this on safari for me? It achieves the same result on Chrome, and I think it might fix the issue for you.

In this line:
https://github.com/daattali/beautiful-jekyll/blob/e8de7726c53c27095526be30dcbe2aef65676803/assets/css/beautifuljekyll.css#L314

**Replace** the margin-top with `bottom: -1.5rem`

And in this line:
https://github.com/daattali/beautiful-jekyll/blob/e8de7726c53c27095526be30dcbe2aef65676803/assets/css/beautifuljekyll.css#L351

**Add** a rule `bottom: -1.9375rem`


And I added you to 
https://beautifuljekyll.com/examples/

lexicalunit:
Just checked and that tweak works perfectly for me on macOS Safari, Firefox, and Chrome 🎉

daattali:
Awesome :) I'm going to hold off for a few days until I can fully test it, but will probably use this fix

Issue #770, opened 2021-01-29
Feature suggestion: Blog post search bar


EthanMcBloxxer:
A search bar to more easily navigate blog posts could be beneficial when having 4+ pages.
A possible candidate for this could be [Simple-Jekyll-Search](https://github.com/christian-fei/Simple-Jekyll-Search). I've implemented it on my website already and think it would be a good addition to Beautiful Jekyll.

[JSFiddle Demonstration](https://jsfiddle.net/d1h79vsf/) (this is done with liquid and a "search.json" config file, so no results are displayed on the example)
[Live Demonstration](https://ethanmcbloxxer.github.io/beautiful-jekyll/) (a font would be good too)

I already have a forked version of Beautiful Jekyll with this, so I could also submit a PR.

daattali:
Thanks for the suggestion Ethan. It's definitely a nice feature, but I generally try to avoid feature creep and keep the theme simple. If this feature can be incorporated with minimal changes and is very simple for the user to turn on/off - eg. by simply switching a config parameter, then I'd be happy to merge

EthanMcBloxxer:
My strategy was to create "search.html" in includes, then include it in the home layout if `post_search: true` in _config.yml.

daattali:
Done in commit 950c341739a876b0f73de23e59839b42e8869b69

I added a search button to the navbar so that the search will be available on every page and so that it will work even for people who don't use the default `layout: home` homepage, and so that it doesn't take up space in the middle of the page (it didn't seem right to me)

On large screens there's a magnifying glass icon at the end of the navbar:

![image](https://user-images.githubusercontent.com/952340/120118269-98877480-c15f-11eb-9c3f-af5410da387d.png)

On smaller screens with the collapsible menu, that button says "Search":

![image](https://user-images.githubusercontent.com/952340/120118281-ac32db00-c15f-11eb-8cc0-96e082710986.png)

In both cases, an overlay is created that can be closed either by clicking the X button inside the input or the Escape key on the keyboard:

![image](https://user-images.githubusercontent.com/952340/120118329-00d65600-c160-11eb-9e10-9a23366e9ec9.png)

PR #771, opened 2021-01-29
Blog post search bar

EthanMcBloxxer:
#770 
This implements simple-jekyll-search, a library that indexes posts and returns them as search results.
(Warning: the fork has the `post_search` property to true by default)


daattali:
Thanks for the PR!
Is there a way to not require an extra json file? And did you ensure the input adapts to the user's colour scheme (defined in _config.yml)

EthanMcBloxxer:
> Is there a way to not require an extra json file?

I don't think so, [the setup of Simple-Jekyll-Search](https://github.com/christian-fei/Simple-Jekyll-Search#getting-started) states the requirement of an extra json file, but it might be possible to host it on a separate server if that would help.

> And did you ensure the input adapts to the user's colour scheme

It now mirrors `site.text-col` (with a default of `#757575`)

daattali:
I prefer not to add anything that requires special files, because there are many feature requests being made continuously and adding a file per feature would add a lot of bloat. Another reason is that anyone using the theme as a gem rather than by forking would have to copy this file to their directory.

If the author of the search plugin can provide a way to not require an additional file - if there can be a default or if it can be included inside the javascript somehow - that would be great. If that's not possible, I'm afraid I prefer simplicity over more features.

rupeshtiwari:
Hi I am waiting for search feature to come in. Really excited to see when this is resolved and come to us.

daattali:
I understand the feature is useful. However please keep in mind I am a sole maintianer of a project used by tens of thousands of people that is aimed mostly at beginners, so there are other considerations at play as well. "Feature creep" and adding more things that need to be maintained for example. That's why I would add it only if it has minimal impact on the codebase and on end users (adding a non-standard file that must be copied to a new repo would unfortunately violate this).

rupeshtiwari:
Thanks @daattali and I can understand we should follow the repository guidelines. I am new to Ruby. However, I will see if I can help here. Because I love beautiful theme and it is lacking with search feature. I want to add this feature.

daattali:
I'm hoping the json can be embedded in the javascript, or that there's a default that doesn't require an external json file

EthanMcBloxxer:
> I'm hoping the json can be embedded in the javascript

The only solution that I could think of right now would be a `data` URI, but it won't work with any version of IE.

daattali:
It would be wonderful if I could drop support for IE11, there are so many things that are held back because of it.

I'm warming up to the idea of having the extra json file is that's the only way to get this working. But instead of the current UI, what do you think of having a magnifying glass icon in the navigation bar so that it's accessible from anywhere?

image

EthanMcBloxxer:
I do quite like this idea (and the icon), but two things (which aren't _exactly_ drawbacks):

1. The search only looks through blog posts in the first place (this can be overridden)
2. You'd also now need a separate `search.html` or to embed it into a popup (that wouldn't be too bad)

daattali:
I would vote in favour of the search being site-wide rather than only in blog posts by default, although it's not a very strong opinion.

I do have a stronger opinion about not including an additional html file, because (1) Any additional files in the top-level make the organization of the repo less clear and intimidates beginners (who are the primary target audience) and (2) For non-beginners who choose to go the ruby gem route, this is an additional file they have to copy that doesn't get included automatically. The `_includes/search.html` file is fine because it's hidden from users and gets automatically included when using the gem version. A not-super-clean solution would be to have the search popup be generated directly from javascript.

EthanMcBloxxer:
I just found out there is a way to trash `search.json`, thanks to me being very new to JS I completely missed the possibility of `JSON.parse`. I hope I didn't mess up the commit history (my repo is 2 commits behind and I don't know much about what to do there, do I delete my fork and refork or pull and merge?). I also don't know enough about JS to make a navbar popup, and too little about liquid to make the `for` loop index all pages.

daattali:
That's great about the json!  As long as the new commits don't directly interfere with your changes, you don't need to worry about it.

I'm going to hold off on accepting this PR because I do want to see if it can be done from the navbar instead of adding it to the homepage. I'm very caught up in several other projects ATM so I can't dedicate the time to look into it, but I would appreciate if someone else can try to tackle it

daattali:
Thanks a lot for all your work on this @EthanMcBloxxer . I'm going to close this PR because I was able to take your idea and implement it into the navbar

PR #780, opened 2021-02-05
feat(footer-scripts): Adding anchors feature for the site

rupeshtiwari:
I want to enable the anchors on the site.  By default I kept it disabled. User can enable it from `_config.yml` file. This is how it will look like: 
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,16 +13,30 @@
 
 ## Table of contents
 
+- [Beautiful Jekyll](#beautiful-jekyll)
+  - [Table of contents](#table-of-contents)
 - [Features](#features)
-- [Sponsors 🏆](#sponsors)
-- [**Build your website in 3 steps**](#build-your-website-in-3-steps)
+- [Build your website in 3 steps](#build-your-website-in-3-steps)
+  - [The easy way (recommended!)](#the-easy-way-recommended)
+    - [1. Fork this project](#1-fork-this-project)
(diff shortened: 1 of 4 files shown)


EthanMcBloxxer:
If you're going to implement AnchorJS I recommend adding more customizable parameters than just toggling it.

<details><summary>Show more</summary>

such as

```js
anchors.options.placement = '{{ site.anchors.position }}';

{% if site.anchors.hover == true %} 
anchors.options.visible = 'hover'; 
{% endif %}

anchors.options.icon = '{{ site.anchors.icon }}';
anchors.options.ariaLabel = '';
anchors.options.titleText = 'Heading Permalink';
anchors.add('{% if site.anchors.elements %}{{ site.anchors.elements }}{% else %}h1, h2, h3, h4, h5, h6{% endif %}');
```

and say the following YAML:

```yaml
anchors:
  placement: left
  hover: true
  icon: "#"
```

would yield the following result:



and with some quick styling:

```css
.anchorjs-link {
  transition: all .25s linear;
  color: #ff8861;
}
.anchorjs-link:hover {
  text-decoration: none;
}
*:hover > .anchorjs-link {
  margin-left: -1.125em !important;
}
```

you could also make this:



in general, I also recommend only loading it on posts (since tags on post previews are a bit odd), but you could always use `anchors.remove(.post-title, .post-subtitle);` or `anchors.add(h2:not(.post-title));` (for just the post title)

</details>

daattali:
Thanks for the initiative @rupeshtiwari 

I don't think I'll include this feature, but I really appreciate your contribution!

PR #788, opened 2021-03-11
Itch.io social link

Emaleth:
Added Itch.io to social links


daattali:
Have you tested this to make sure the fontawesome version used in this theme has this icon?

Please also add an entry in the CHANGELOG file

Emaleth:
Added entry in CHANGELOG.md (under unreleased) but cant find FA in head or anywhere, could you tell me where to find info / tell what version you use?

Emaleth:
Ok, found it. Yes, it is in version 5.

daattali:
Have you tested this?

Emaleth:
https://emaleth.dev
https://github.com/emaleth/emaleth.github.io
(keep in mind it's just social include, its still of original repo via remote_theme, but since other files were not modified it should suffice.

daattali:
Thanks!

PR #799, opened 2021-04-16
use base_url as default when absolute is null

randellma:
When doing `jekyll build` (not serve) and the site.url and site.base_url are not configured, the home links would resolve as `href=""` (empty) and instead of returning home they would refresh the page instead.

Switching to `{{ '/' base_url | absolute_url }}">` will make them default to `href="/"` or `href="/[base_url]"` appropriately. I can't think of a situation where an empty href would be desired instead.

This change does not affect github pages and jekyll serve as the `absolute_url ` will not be null in those situations.

### DIFF
--- a/_includes/footer-minimal.html
+++ b/_includes/footer-minimal.html
@@ -12,7 +12,7 @@
 
     {% if site.url-pretty %}
       &nbsp;&nbsp;&bull;&nbsp;&nbsp;
-      <a href="{{ '/' base_url | absolute_url }}">{{ site.url-pretty }}</a>
+      <a href="{{ '/' | absolute_url }}">{{ site.url-pretty }}</a>
     {% endif %}
   </div>
 </footer>
--- a/_includes/footer-minimal.html
+++ b/_includes/footer-minimal.html
@@ -12,7 +12,7 @@
 
     {% if site.url-pretty %}
       &nbsp;&nbsp;&bull;&nbsp;&nbsp;
-      <a href="{{ '' | absolute_url }}">{{ site.url-pretty }}</a>
+      <a href="{{ '/' base_url | absolute_url }}">{{ site.url-pretty }}</a>
     {% endif %}
   </div>
 </footer>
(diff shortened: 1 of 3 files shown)


daattali:
To be honest this kind of change makes me very nervous, as I'll have to test it out in many different scenarios. Was the problem **only** with the empty home link? If so, I would prefer changing the home link rather than changing the way links are resolved

randellma:
Yeah fair enough. Changing only the home links is what I was trying to do. There's just a couple instances of them in the nav and footer.

I updated the PR to to something less drastic and more appropriate that still works.

Really the issue is that the absolute_url filter doesn't work if you don't have both url and base_url configured.

However, maybe the best solution is to just include a note in the README to say that if using `jekyll build` that you _must_ have both `url` and `base_url` in the configuration file.

randellma:
Thinking about it a bit - what I'm doing here is really just a workaround for misconfiguring jekyll.

daattali:
So these two fields are required only for `jekyll build`? Is there documentation anywhere that explains this clearly?

And if there are instances in the theme currently that use "" as a home link, would changing those to "/" be helpful?

randellma:
It's more that they are required for the absolue_path liquid filter. And jekyll build does not supply a site.url unless configured whereas jekyll serve supplies "0.0.0.0" and github pages somehow automatically supplies the correct [username].github.io.

> site.url
> Contains the url of your site as it is configured in the _config.yml. For example, if you have url: http://mysite.com in your configuration file, then it will be accessible in Liquid as site.url. For the development environment there is an exception, **if you are running jekyll serve in a development environment site.url will be set to the value of host, port, and SSL-related options. This defaults to url: http://localhost:4000.**

So absolute_url being used like {{ '' | absolute_url }} will resolve to a blank string if site.url is missing when running jekyll build.

But yeah - changing links to the home page to "/" would work but thinking about it more there could be other problems for any other place that absolute_url is used and not expected to be blank.

daattali:
From that quoted paragraph, they say "in a development environment". Is this really only an issue for development?

daattali:
Thanks for all your feedback. I'd be happy to merge a PR that adds a line about the url when using 'jekyll serve'

PR #814, opened 2021-05-15
Restore feed link as a social network link

kzu:
In previous versions of the theme, the feed link would be alongside the other social links, [for example](https://cazzulino.com):

image

This PR makes it possible (again) to enable it with:

```
share-links-active:
  feed: true
```

This aligns with the new mechanism for enabling those icons using the new `share-links-active` setting. 

See it in use at https://anycoder.net with the [config shown above](https://github.com/anycarvallo/anycarvallo.github.io/blob/main/_config.yml#L46).

### DIFF
--- a/_includes/social-networks-links.html
+++ b/_includes/social-networks-links.html
@@ -1,6 +1,6 @@
 <ul class="list-inline text-center footer-links">
 
-{%- if site.rss-description -%}
+{%- if site.rss-description and site.share-links-active.rss -%}
   <li class="list-inline-item">
     <a href="{{ '/feed.xml' | relative_url }}" title="RSS">
       <span class="fa-stack fa-lg" aria-hidden="true">


daattali:
You can do this with the `rss-description` value in the config value, no? https://github.com/daattali/beautiful-jekyll/blob/a531737107167728c5fc8e25a4aadbd65d47533c/_config.yml#L66-L68

kzu:
It escapes me how setting a *description* would turn on a footer link, but ok, I guess?

kzu:
Ok, looked at the source. I still think it would be more intuitive to check for either an rss-description (for backs compat, since the actual description isn't used at all in that footer snippet?) but also check for a boolean `share-links-active.feed`, which seems more natural to me. I can update the PR accordingly if you think it's worth it.

daattali:
Thanks for your input. I admit the RSS is a bit awkward because it doesn't fit quite nicely into any other category.

I don't think `share-links-active` is the right place for rss, because that list is responsible for creating buttons to share the current page on social media platforms. 

I removed it from `social-network-links` because all other parameters there take a unique value, but RSS was the odd one out by just requiring a `true` value. So it felt strange to have that parameter there.

After spending some time thinking about it, I think this might be the cleanest solution: right now, an RSS feed is [always added to every page's head](https://github.com/daattali/beautiful-jekyll/blob/99fdbbd59354193e51525910eee09c855512e176/_includes/head.html#L42), and setting `rss-description` adds a link to the footer. Instead, we can make it so that setting `rss-description` causes the link to get added to the page, and if you also set `rss: true` in `social-networks-links` then it also adds it to the footer. If `rss: true` is set but `rss-description` is not set, nothing happens. This behaviour will be documented in the `rss-description` parameter.

If that makes sense, I'd be happy for a PR.

kzu:
Any particular reason why you call it "rss" everywhere when in reality he can also be atom? Why not just "feed" instead?

Other than that, it sounds good, yes. I can update the PR, sure

daattali:
Just because RSS is more widely known/more of a household name. Atom is known to web developers but it is a more niche term in my opinion

kzu:
That's an accurate perception, I think, but it still leaves an atom-formatted feed named `rss` in the current template :/.

kzu:
Ok, made the change, let me know what you think!

daattali:
> we can make it so that setting rss-description causes the link to get added to the page, and if you also set rss: true in social-networks-links then it also adds it to the footer. If rss: true is set but rss-description is not set, nothing happens. This behaviour will be documented in the rss-description parameter.

There are 4 things that need to happen:

- the rss link gets added to the page only when `rss-description` is available
- setting `rss-description` and `social-networks-link.rss: true` adds an icon to the footer (you did this)
- the `social-networks-link.rss` parameter needs to be added to the config file so that users know it exists
- the documentation for `rss-description` in the config file should explain this

This PR needs to address all these sub tasks to be merged :)

daattali:
@kzu I plan on releasing a new version soon, so let me know if you want to include this PR

kzu:
nevermind. I'll just maintain my fork.

daattali:
Reopening because I do agree with the idea of this PR

kzu:
Here's the version I'm using, which is more natural to me: https://github.com/kzu/beautiful-jekyll/commit/86d447a82c17d6d65da5426077a1fdb8a443ca86

PR #815, opened 2021-05-16
Include Reddit in share options of posts

### LINKED ISSUE #754: Include Reddit in share options of posts
Please add Reddit as a share option that appears below the pages marked as a post and added in the **_post** folder.

I am requesting you to add the option for posts to be shared on Reddit.

Users should be able to configure the `_config.yml` file in such a manner-

```yaml
# --- General options --- #

# Select which social network share links to show in posts

share-links-active:
  reddit: true # add this option
```
and an icon is added below every post so that readers can share the post to Reddit.

l33d4n:
Closes #754


daattali:
Thanks. Can you please update the CHANGELOG

PR #848, opened 2021-08-09
Added feature to customize hover links color in footer

DoctorD90:
Hello! I just added the feature to customize the hover/focus links color in the footer.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -109,6 +109,7 @@ hover-col: "#0085A1"
 footer-col: "#EAEAEA"
 footer-text-col: "#777777"
 footer-link-col: "#404040"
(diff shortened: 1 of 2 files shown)


daattali:
Can you please add an entry to the CHANGELOG

_config.yml
```diff
@@ -108,7 +108,8 @@ link-col: "#008AFF"
 hover-col: "#0085A1"
 footer-col: "#EAEAEA"
 footer-text-col: "#777777"
-footer-link-col: "#404040"
+footer-link-col: "#0085A1"
+footer-hover-col: "#404040"
```
daattali, line 112: Why did your flip the colours? Won't this cause a behaviour change?

PR #884, opened 2021-10-03
Build page locally with `nix`

fkarg:
This PR allows for locally building GitHub pages with `jekyll` and `nix`. It will look exactly like the one GitHub will build.

When having `nix` installed (find out more about nix [here](https://nixos.org/guides/how-nix-works.html) or see the install guide [here](https://nixos.org/guides/install-nix.html)), running `nix-shell` builds the `shell.nix`, which (locally, think virtual environment) installs `jekyll`, `jekyll-paginate` and starts the `jekyll`-server in incremental build mode.

Initial page generation takes about two seconds, incremental changes to files (`*.md`, `*.html`, `*.js`, `*.css`, ...), take about 0.2sec (on my 6yo laptop) to rebuild. Changes to `_config.yml` require a deletion of `_site` artifacts and restart.

 I'd appreciate a `hacktoberfest-accepted-label`. :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,7 @@ ehthumbs.db
 Gemfile.lock
 
 beautiful-jekyll-theme-*.gem
+
+# local build
+.jekyll-cache/
+.jekyll-metadata
--- a/README.md
+++ b/README.md
@@ -204,6 +204,18 @@ Visit the [Official website](http://beautifuljekyll.com/examples) to see sample 
 
 If you'd like to showcase yourself and join this list, [upgrading to the Individual plan](https://github.com/sponsors/daattali/sponsorships?&tier_id=7362) will give you that publicity plus some other rewards!
 
+# Build the site locally
+Locally building the site allows you to see changes before commiting them. This enables much faster iteration cycles for small or visual changes (think changes to css or colors). GitHub pages uses `jekyll` to build the page in `_site`.
+
+If you have `nix` installed (if not, see [here](https://nixos.org/guides/how-nix-works.html) what it is and how it works, and [here](https://nixos.org/guides/install-nix.html) for installation), you can build this page locally simply via executing `nix-shell`.
+
+Running `nix-shell` builds the `shell.nix`, which (locally, think virtual environment) installs `jekyll`, `jekyll-paginate` and starts the `jekyll`-server in incremental build mode. You can find your locally built page on `localhost:4000`.
+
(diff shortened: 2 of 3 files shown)


fkarg:
Sure!

First, nothing is installed globally, contrary to what `sudo apt-get install` does. Depending on familiarity/usage of ruby/gems/bundler, this might lead to version conflicts or installed, but unused packages (admittedly, neither of these are a major issue). `nix-shell` creates a virtual environment dedicated for local development with nothing else in it, and is fully reproducible. This advantage isn't exactly demonstrated here, so it might not be the main reason to include this.

Second, for users who already use nix, this is the fastest and easiest way to get started. `shell.nix` and `default.nix` are filenames most people using `nix` are relieved to see - it'll work out of the box, in one simple step (just starting the `nix-shell`). On `NixOS` you would need to create one anyways (if you don't want to install it globally on your system or user), and even when `apt` is available, it allows you to use all advantages of virtual environments.

Adding plugins and making changes is rather straightforward. It's also possible to [build the gems individually](http://stesie.github.io/2016/08/nixos-github-pages-env) for more complex ruby setups. In my experience, this is noticably slower though.


Sidenote: I was surprised to see the installation instructions, and I've been using this theme for two years now. Even [in the readme](https://github.com/daattali/beautiful-jekyll#the-hard-way-using-ruby-gems), it's not immediately obvious that this would enable me to build the page locally. I added a note about that in a commit just now, and I'd create a separate PR if you don't want to merge this one. Especially after having set it all up, my place to look for information about that would be the readme, not the original page.

daattali:
Thanks for your feedback. There is a small section in the README ([here - " skip the easy way"](https://github.com/daattali/beautiful-jekyll#build-your-website-in-3-steps) and [here](https://github.com/daattali/beautiful-jekyll#the-hard-way-using-ruby-gems)) but you're right that it's probably not easy enough to notice. I'll take your suggestion to make that clearer.

Regarding adding nix, I'm very conflicted about this. I've done some research and from what I can tell it's not a very popular tool. It seems very useful and loved by its users, but it's just not widely adopted. I've made the conscious decision a long time ago that I want to try to cater to the largest audience and make this theme simple and uncluttered, so I often have to reject features that are beneficial for a select few in order to keep the simplicity down. There used to many more files in the root level of this repo, with all sorts of configuration files for different tools, which caused a lot of confusion to many users, and this is that path I want to avoid going down again. 

I won't be merging this, but I do want to thank you for the suggestion and the initiative. I hope you understand

PR #888, opened 2021-10-06
add giscus support

### LINKED ISSUE #886: giscus support
I'm reaching the finishing stages of adding [giscus](https://giscus.app) comments support for my Beautiful Jekyll site. If you're not familiar, giscus is what I would consider the "spiritual successor" to utterances as it uses GitHub Discussions rather than GitHub Issues for its comments.

Do you have any interest in me putting together a PR to add that functionality? Setup is almost identical to utterances so I don't think it adds much, if any, complexity (apart from a slightly longer config file).

patgarz:
Adds support for [giscus](https://giscus.app) comments.

Essentially a clone of the current setup for utterances, but with the config values switched out for giscus.

closes #886


daattali:
@patgarz  I'm planning on releasing a new version soon. Can you please show me a working example of this feature?

I tried it in a test site https://daattalitest.github.io/ but it doesn't seem to work - here's the config  https://github.com/daattalitest/daattalitest.github.io/blob/b8266d2af5538dc18d4621bd65145d1a55701685/_config.yml#L191-L200
 
I keep getting 404's from discus.app so nothing happens when I try to submit a comment.

PR #907, opened 2021-11-29
Add discord to supported social networks

emhl:
This adds the option to display a discord icon in the social network list.

you could either set this to for example invite/invite_code to get a link to your discord server or to users/userid to get a link to your discord profile where people could add you

apperantly this theme uses an old version of fontawesome and because of that it still displays the old discord logo


daattali:
Thanks!

PR #911, opened 2021-12-06
Updated Discord Link from .com to .gg

Vombato:
I don't know since when but new Discord invites use the `.gg` domain instead of `.com` one.

### DIFF
--- a/_includes/social-networks-links.html
+++ b/_includes/social-networks-links.html
@@ -302,7 +302,7 @@
   
 {%- if site.social-network-links.discord -%}
   <li class="list-inline-item">
-    <a href="https://discord.com/{{ site.social-network-links.discord }}" title="Discord">
+    <a href="https://discord.gg/{{ site.social-network-links.discord }}" title="Discord">
       <span class="fa-stack fa-lg" aria-hidden="true">
         <i class="fas fa-circle fa-stack-2x"></i>
         <i class="fab fa-discord fa-stack-1x fa-inverse"></i>


daattali:
Is it backwards compatible? Do both versions work?

Vombato:
It seems not, at least for the invites I create 😅

daattali:
I'm not very familiar with discord and how these URLs work with them. My username is "deanat78" but both .gg and .com don't seem to work with my username. Are you sure these URLs work?

Vombato:
> I'm not very familiar with discord and how these URLs work with them. My username is "deanat78" but both .gg and .com don't seem to work with my username. Are you sure these URLs work?

Actually to hava link to a profile you need to use the `https://discordapp.com/users/xxxx/` where `xxxx` stands for the user's ID, not the username. You can obtain an user's ID by activating developer mode, eg: https://discordapp.com/users/539800670652465193/ should direct to my profile.
I was initially referring only to the invite system which works with `https://discordapp.gg/xxxx` where`xxxx` is the invite ID that you generate when creating an invite link to a server/channel. Also, Discord's partnered servers can use a custom unique server name which sobstitute the ID.
I didn't see that the user that implemented the Discord feature in this repo also included "username" in the suggestions, if it's ok, later I'll make a PR fixing that too so that it would be more clear.

Emily2010:
这是来自QQ邮箱的假期自动回复邮件。
邮件已收到

daattali:
> I didn't watch carefully that the user that implemented the Discord feature in this repo also included in the comments:
# discord: invite/invite_code or users/userid

Right, I just noticed that too. No bug then :)

PR #912, opened 2021-12-06
Removed fixed thumbnail size specification

kenjinote:
When displaying thumbnails on a mobile device, it is displayed small.


daattali:
It is meant to be small on mobile. 

image

If it looks way too small for you, please share a screenshot, but keep in mind this is on purpose

PR #940, opened 2022-02-17
fix dropdown item text color

divideconcept:
Fix dropdown item text color in the navigation bar to follow site.navbar-border-col (otherwise it inherits a default color from a bootstrap theme)


daattali:
This line already exists, just a few lines below yours

divideconcept:
Indeed ! No idea why it wasn't working on my computer, maybe I had an old version of the theme.

PR #941, opened 2022-02-17
make LinkedIn link compatible with company pages

divideconcept:
Company pages on LinkedIn have urls such as https://www.linkedin.com/company/mycompany.
Removing /in/ allow such pages to be linked with beautiful jekyll

### DIFF
--- a/_includes/social-networks-links.html
+++ b/_includes/social-networks-links.html
@@ -98,7 +98,7 @@
 
 {%- if site.social-network-links.linkedin -%}
   <li class="list-inline-item">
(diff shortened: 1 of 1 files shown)


daattali:
This will break existing links, won't in? For example if I currently have "daattali" in the config file then right now the URL works (https://www.linkedin.com/in/daattali) but after this change it won't (https://linkedin.com/daattali)

divideconcept:
Yes unfortunately it will.
As a workaround if we keep the code as is, users can set the link to "/../company/blablah"

daattali:
I suppose more flexibility is better, so it would be a good idea to remove the "/in"

Could you make all the related changes, such as updating the config file, the changelog, and anywhere else that might mention LinkedIn

PR #950, opened 2022-03-14
Markup home page content lists and tags as lists

JeffersonBledsoe:
This PR wraps the posts list on the home page in a list tag to fix it's semantics. The same fix has been applied to the list of tags for a post.

### DIFF
--- a/_layouts/home.html
+++ b/_layouts/home.html
@@ -9,8 +9,8 @@ layout: page
 <!-- role="list" needed so that `list-style: none` in Safari doesn't remove the list semantics -->
 <ul class="posts-list list-unstyled" role="list">
   {% for post in posts %}
-  <li>
-    <article class="post-preview">
+  <li class="post-preview">
+    <article>
 
       {%- capture thumbnail -%}
         {% if post.thumbnail-img %}


daattali:
Thanks, do you have a version of beautifuljekyll that uses this exact PR so that I can test it visually?

JeffersonBledsoe:
Hi @daattali. I'm not personally using this theme but know somebody who was looking into it. Here's a couple of screenshots showing the before & after of the change using the example content in this repo. The visual changes are minimal (and can be completely removed if desired).

## Before

Screenshot of the beautiful-jekyll theme without the new markup


## After

Screenshot of the beautiful-jekyll theme showing a difference in list separator styling with the new markup

daattali:
Other than adding `role="list"` in order to fix what you mentioned on Safari, is it necessary to make the other changes of adding ul/li tags?

JeffersonBledsoe:
@daattali The markup makes it clearer for non-visual agents (such as screen readers, SEO engines or bots) what the markup is. For example, screen readers will announce the user is in a list and provide functionality for navigating lists. I need to do some more research on whether the tags should be in a `<ul>`, but the list of pages almost certainly should be

daattali:
I would need to do some rigorous testing on this to make sure it doesn't affect the UI. Have you tested this? Also, there's a typo `ole` instead of `role`

daattali:
I've made a few changes to make usre the styling is not changed, merging now. Thanks!

PR #976, opened 2022-04-30
added hackerrank social link

tanersekmen:
added hackerrank social link

There was no information about hackerrank and I noticed that and then I added hackerrank social link


tanersekmen:
I added hackerrank social link due to there was nothing about it.

daattali:
This won't actually do anything, you'll need to modify the code to actually add the icon in the correct place. You can look at previous PRs from other people who added different social media icons to see how to do that

tanersekmen:
I added social media networks and opened new PR. Can you check that? Let me know if something wrong.

PR #985, opened 2022-05-18
footer-extra error

danielp123:
footer-extra was using page.footer-extra instead of site.footer-extra and was not working. Changed now to correct bug.

Please note that if you are trying to update **your** website, this is the wrong place to do so. Please carefully follow the Beautiful Jekyll instructions (found at https://github.com/daattali/beautiful-jekyll#readme) and make sure you submit changes to **your** version of the project.

If your intention is to submit a Pull Request, please describe what your pull request achieves.

Thank you!

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -3,9 +3,9 @@
     <div class="row">
       <div class="col-xl-8 offset-xl-2 col-lg-10 offset-lg-1">
       {% include social-networks-links.html %}
-      {% if page.footer-extra %}
+      {% if site.footer-extra %}
         <div class="footer-custom-content">
-          {% for file in page.footer-extra %}
+          {% for file in site.footer-extra %}
             {% include {{ file }} %}
           {% endfor %}
         </div>


daattali:
It's correct, it's not a bug. `footer-extra`, like many other YAML parameters, are per-page. This way you have the flexibility of using it only on one or some pages, not on the entire site. You can use default values in _config.yml if you want to set it for the entire site

danielp123:
@daattali Thanks for the explanation, but I couldn't get that to work. I put in "footer-extra: footer-disclaimer.html" in my _config.yml file and added the extra footer-disclaimer.html in my _includes folder, but nothing happened.
Do I need to do something else in order to get the footer-extra on each page?

daattali:
That's not how default values work. Read the documentation on jekyll default values: https://jekyllrb.com/docs/configuration/front-matter-defaults/

daattali:
I can't read it, but it does look very professional

Issue #1004, opened 2022-07-07
Feature request: Edit in GitHub link/button


dagnelies:
Hi, 

First, thanks for this beautiful theme. The only thing I miss is a "Edit in GitHub" link or button on the pages. This would be a nice addition.

daattali:
That's a good idea. I'll have to think of a way to incorporate that into the UI in a sensible way and take into account that not all usages of beautiful-jekyll are on github.

daattali:
It's not perfect though, if anyone wants to improve it, I'd be happy to accept PRs. For example, if you're on a page that doesn't actually exist on github, it'll still show this button but take you to a 404 page. For example, if you navigate to the second page of the blog/feed, it'll try to go to a page called "page2/index.html" in your github repo.

dagnelies:
Amazing 😲  🎉 I didn't expected you to implement that so quickly, impressive!
I'm gonna try it out right now.

dagnelies:
Nice, works great, thanks!

PR #1006, opened 2022-07-14
Add "New Page" option for 404

sorat0mo:
Improvement on #1004

This should work, but it cannot distinguish new pages from new posts, i.e. When you try "abc.github.io/2022-07-14-abcde" it will just direct you to create a file called "2022-07-14-abcde" in root.

Example when the page does not exist (Notice the text in red boxes)

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -35,7 +35,7 @@
         &nbsp;&bull;&nbsp;
         {% if page.path == '404.html' %}
           <script>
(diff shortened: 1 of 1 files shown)


daattali:
Thanks @sorat0mo , I appreciate the effort. But for the sake of not bloating the codebase with too many esoteric features, I prefer not to merge this feature. "Edit this page" is something that can be used by anyone who goes to your website when they see an issue and they want to help you fix it; "New page" button is only for yourself and it sounds like a bit of a strange use case. I prefer not to add code/features that are not going to be very popular.

PR #1010, opened 2022-07-19
Fixes #765

### LINKED ISSUE #765: Feature suggestion: Menu width based on the width of the longest word
Hi Dean,

After asking on Twitter about this, I'm creating this "issue" with a feature suggestion.

Menu width should be based on the width of the longest word in menu items.

If not, words are "truncated" to the next line. For example, a menu called "ABOUT" with 3 items ("ME", "LAPTOP", "SOFTWARE") is rendered like this:

image

dagnelies:
Fixes #765, I also had the same issue.

Before:

image

After:

image

I also restored the small border since I found it nicer. It's up to you if you wanna leave it or not.

Works also responsive.
See https://dagnelies.github.io/passwordless-www/ for live demo.

### DIFF
--- a/_includes/nav.html
+++ b/_includes/nav.html
@@ -16,7 +16,7 @@
         {%- if link[1].first %}
           <li class="nav-item dropdown">
             <a class="nav-link dropdown-toggle" href="#" id="navbarDropdown" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">{{ link[0] }}</a>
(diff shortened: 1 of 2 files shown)


daattali:
Thank you very much! I made some changes, but merging

daattali:
I made the default behaviour go back to the original because it is often useful. The "fix" where sub menu is longer will now happen when you set `navbar-var-length: true` in the config.

PR #1115, opened 2023-03-20
Allow html in share tags

scotch83:
This PR aims to allow use of html in the share-description and share-title yaml of the pages, to allow for styling in web search results and web page sharing.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -4,7 +4,7 @@
 
   {% capture title %}
     {%- if page.share-title -%}
-      {{ page.share-title | strip_html | xml_escape }}
+      {{ page.share-title | xml_escape }}
     {%- elsif page.title -%}
       {{ page.title | strip_html | xml_escape  }}
     {%- else -%}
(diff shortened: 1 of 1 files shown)


flyingotter (approved):

daattali:
Interesting, I didn't know that they can include HTML, I thought only plain text was supported in search engines. Correct me if I'm wrong, but this will have more consequences than simply styling on search engines, this will also affect the browser tab title, twitter/facebook/social media share cards, and RSS title -- and not all of them support HTML tags. Is that correct?

scotch83:
@daattali I am sorry but I cannot say it with certainty as I did not test all, but I found myself all this out by testing my website on metatags.io and start fiddling with styling tags. It seems it's reflected, at least for Facebook, Twitter and Google results.
I would assume they all rely on the same sanitisation and possibilities, and if they don't they are not implementing the feature. Somewhere I have even seen people using even specific font families for titles. I believe it is anyways an open field to leave the users with freedom of choice. I am glad you guys have approved it 🙂🙏

daattali:
@scotch83 I appreciate your initiative in making this PR, but unfortunately I don't think this is a good idea because of the reasons I mentioned previously. This small change will have several other consequences. You may do this in your own repository, but I don't want to cause this behaviour to everyone else.

For example, here is what I meant by the browser tab title being affected:

image


And here is what I meant about facebook  share card looking broken:

image


In both cases, the HTML tag is being shown (they cannot be **rendered** as HTML), that's why the code is stripping out tags.

scotch83:
Thanks for your attention and feedback anyways, sorry for the disturb and time waste.

PR #1129, opened 2023-04-20
Correct discord social link

NicoIsLost:
The discord social link for servers is discord.gg, not discord.com. The current version does not link to discord servers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 - Added `footer-hover-col` config setting to customize the hover colour of links in the footer (#848)
 - Added social network links for Discord (#907), Kaggle (#961), and Hackerrank (#978)
 - Made the home page feed more accessible for screen readers (#950) 
+- When using Google Analytics config setting, encourage using the new `gtag` and discourage using the deprecated `google_analytics` (#1096)
 
 ## v5.0.0 (2020-09-15)
 
--- a/_config.yml
+++ b/_config.yml
@@ -159,11 +159,8 @@ footer-hover-col: "#0085A1"
 # --- Web Analytics Section --- #
 #################################
 
-# Fill in your Google Analytics gtag.js ID to track your website using gtag
-#gtag: ""
-
-# Fill in your Google Analytics ID to track your website using Google Analytics
-#google_analytics: ""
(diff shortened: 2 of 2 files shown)


daattali:
I don't use discord so I'm not 100% about this, but doesn't the current version work for discord users? Is the URL for users vs servers different?

NicoIsLost:
I've never seen a discord link for a user, but it looks like it is possible. The server one I would think is most important. 

Here's instructions for the user link: https://www.followchain.org/discord-profile-link/

I just tried it and it works, so you're right, but it's different formatting, so perhaps two options is best?

daattali:
I'm inclined not to support two different links for the same service, because that could get confusing. As far as I know, discord only has one icon, it's not two platforms with different icons and different websites for users vs servers. I'll have to look into this.

daattali:
The current config file shows this line for discord:

```
discord: invite/invite_code or users/userid 
```

It looks like the `invite/invide_code` is actually what you want. If you go to `discord.gg/test` it redirects to `discord.com/invite/test`. So we CAN support both, both user and servers actually do go to discord.com! 

Right now you can simply use `invite/g9wJSmtnF4` in your own config in order to get it to do what you want.

My guess is that in the past this was the only way to link to a server, and at some point dicord decided to add a new simpler URL that does a redirect.

I don't want to stop supporting the existing formats, but what we can do, if you're willing to take on this task, is:

1. Change the config line to

     ```
    discord: "invite_code" or "users/userid" or "invite/invite_code" 
    ```

2. If the discord parameter starts with the string `users/` or `invite/` , then the link is `discord.com/<config.discord>`. If the discord parameter does not start with one of those two strings, then the link is `discord.com/invite/<config.discord>`. 

Does that make sense?

NicoIsLost:
Yeah makes perfect sense. 

I'll take a look when I get a chance :-)

daattali:
@NicoIsLost do you think you'll get around to this or should we close the PR?

NicoIsLost:
I will look at it, just been traveling this week. You can close and I can do another PR if you don't want it lingering for a while as it might be a little while before I get time...

daattali:
Don't mind it hanging around, just doing some cleanup and wanted to make sure anything that will not get done doesn't stay open

PR #1135, opened 2023-05-11
Updated Discord Link

NicoIsLost:
Per our earlier conversation, the discord link now will work if it is populated with "users/code", "invite/code", or "code".

I hope this is helpful.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -73,7 +73,7 @@ social-network-links:
 #  mastodon: instance.url/@username
 #  ORCID: your ORCID ID
 #  google-scholar: your google scholar
-#  discord: invite/invite_code or users/userid 
+#  discord: "invite_code" or "users/userid" or "invite/invite_code" 
 #  kaggle: yourname
 #  hackerrank: yourname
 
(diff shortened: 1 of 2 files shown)


daattali:
This code only checks if the string contains "users" or "invite" anywhere in it. I prefer to be more precise to reduce the likelihood of a bug.

Something like this is probably safer:

```
{% assign firstpart = site.social-network-links.discord | split: "/" | first %}
{% if firstpart != site.social-network-links.discord and firstpart == "users" or firstpart == "invite" %} the param starts with users/ or invite/ {% else %} it doesn't {% endif %}
```

(I haven't tested the code above but something like that should work)

NicoIsLost:
That should work like you mentioned, we can split using the /, but I was thinking that if they add for instance "invite/sdadasda/" it would still be a valid link, but I am not sure how the split command would handle the second backslash.

Sure there is still the possibility of "invit" or "users" being in the discord code, but I guess the probability is very low, and depending on how they are generated maybe impossible?

daattali:
`split: "/" | first` would only look at the first part. Something like `invite/foo/` or even `invite/foo/bar/` would all still mean that `split: "/" | first` will result in `invite`.

NicoIsLost:
Good to know. Updated it to use split | first. 

Should be good now :-]

daattali:
merged

PR #1168, opened 2023-07-02
Add GitLab to the list of social media sources

agvallejo:
GitLab is also in font-awesome so adding support for it is pretty much a copy-paste from the GitHub entry. I just added it for my own website, but it should probably live upstream.


daattali:
Thanks, I'm surprised this hasn't come up so far. Could you update the CHANGELOG please?

agvallejo:
>I'm surprised this hasn't come up so far
Me too. Folks might be fixing it locally, or simply be unaware it can be done this easily.

> Could you update the CHANGELOG please?
Done 👍

Issue #1171, opened 2023-07-13
Images not working on post pages


natan-dias:
Hello all.

I am not able to put images to my markdown posts, following your example post page. In my case, the code is:

`![diagrama01](/assets/img/post-img/ssh-tunnels-post/diagrama01.jpg)`

But for some reason, my post is only with the "diagrama01" word on it. Is there any way to post images in the markdown posts?

Kind regards.

natan-dias:
Hey, thank you very much. I used the second option and it works like a charm. 

I had been looking for that in your past issues, but I decided to open that, in fact, because I didn't find anything related to that. 

Thanks!!

natan-dias:
I think we are good to close this issue. Thank you very much!

PR #1172, opened 2023-07-13
Feature - Change Readme to add images documentation hint

natan-dias:
Hi.

I took the opportunity to add this hint to your readme file after we discussed via Issues about the Image problem in posts. Feel free to just use this if you want or change it though.

Issue link: https://github.com/daattali/beautiful-jekyll/issues/1171


daattali:
Thanks, I'll think about it for a bit and then decide where I want to document this. I'll also probably want to update the sample posts to show this.

daattali:
I want to keep the README as short as possible and only focused on the forking method, so I'm going to add this info to the FAQ and to the sample markdown post

PR #1218, opened 2023-10-12
Add Bluesky to social network links

chohlasa:
Adds Bluesky as another possible social network link. 

Note I used the same FontAwesome icon (a simple square!) that Bluesky uses on its homepage:
https://blueskyweb.xyz/

The icon is discussed further here:
https://github.com/FortAwesome/Font-Awesome/issues/19810

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,8 @@
 
 - BREAKING CHANGE: Allow changing the order of the social network links that appear in the footer (#1152)
 - Fixed bug where hovering over search results showed the text "{desc}" (#1156) 
-- Added social network link for GitLab (#1168)
+- Added social network links for GitLab, Bluesky (#1168, #1218)
 - Added instructions and example on how to fix image links in project sites (#1171)
-- Added social network link for Bluesky (#1218)
 
 ## v6.0.1 (2023-06-08)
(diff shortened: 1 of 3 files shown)


daattali:
Thanks!

raphacunha:
FontAwesome has released a new bluesky icon:
https://fontawesome.com/icons/bluesky?f=brands&s=solid

Could we have that added to the social network links?

daattali:
It will not work with the current version of fontawesome used in this theme. It looks like it was just released 3 days ago. If you'd like to upgrade the version of fontawesome used in the theme, you can submit a PR for that but please test that all the existing fonts still work and that the UI/UX isn't affected

raphacunha:
I tested it on my website and it seems to be working fine in terms of backward compatibility and UI/UX, but I only use a limited number of fonts, so I can't assure that all existing fonts still work.

daattali:
Please submit a PR so that it can be tested properly

lorddev:
This feature is not working for me. Is there a trick to it?

daattali:
What feature?

lorddev:
This feature: "Add Bluesky to social network links"
I removed ***@***.***” from the yml and it works now.

PR #1220, opened 2023-10-14
Add Support for front matter tag author: to add a by line to posts, w…

sschwetz:
(Resubmitted on correct source branch)

This pull request adds the ability to tag a post with an author's name by using the front matter author: field.

this pull adds/changes 5 files:

_config.yml - adds collection authors; adds scope for _author layout
_authors/stephen.md - example author's file
includes/header.html - adds logic to add link to posts
_layouts/author.html - layout used to generate authors bio pages
_layouts/home.html - adds by line to posts on the home page
_layouts/author.html needs some love to make it look pretty, UIX is not my forte.

### DIFF
--- a/_layouts/home.html
+++ b/_layouts/home.html
@@ -103,21 +103,17 @@ layout: page
   {% if paginator.previous_page %}
   <li class="page-item previous">
     <a class="page-link" href="{{ paginator.previous_page_path | absolute_url }}">
-      <span class="fa-stack fa-lg" aria-hidden="true">
-        <i class="fas fa-circle fa-stack-2x"></i>
-        <i class="fas fa-arrow-left fa-stack-1x fa-inverse" alt="Newer Posts"></i>
-      </span>
-    </a>
+	  <i class="fas fa-arrow-left" alt="Newer Posts"></i>
+	  <span class="d-none d-sm-inline-block">Newer Posts</span>
+	</a>
   </li>
   {% endif %}
   {% if paginator.next_page %}
   <li class="page-item next">
     <a class="page-link" href="{{ paginator.next_page_path | absolute_url }}">
-      <span class="fa-stack fa-lg" aria-hidden="true">
-        <i class="fas fa-circle fa-stack-2x"></i>
-        <i class="fas fa-arrow-right fa-stack-1x fa-inverse" alt="Older Posts"></i>
-      </span>
-    </a>
+      <i class="fas fa-arrow-right" alt="Older Posts"></i>
+	  <span class="d-none d-sm-inline-block">Older Posts</span>
+	</a>
   </li>
   {% endif %}
 </ul>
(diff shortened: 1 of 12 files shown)


_layouts/home.html
```diff
@@ -45,10 +45,15 @@ <h3 class="post-subtitle">
           </h3>
         {% endif %}
       </a>
-
+
```
daattali, line 48: Trim the whitespace

daattali:
Since this PR introduces additional pages that are getting linked, it will require a bit more testing that most simpler features. Have you tested and ensured the links work on: GitHub user page, GitHub project page, local installations?

daattali:
I like the idea of this feature, but I think it needs some work to get it to be production ready.

- I just tested this locally and the links do not work correctly. It tried to go to `http://authors/dean.html`
- Ignore the author on a page if that person doesn't have a matching page in the collection
- If a page has an autho, add the `author` meta tag 
- I looked at many websites and noticed that the standard is the place "By: author" on a separate line, before the date. So let's put it in a line before the date.
- For cleaner UI, the author name colour should match the colour of the rest of the metadata - let's make it **bold** to signify that it's clickable
- Documentation needs to be added

sschwetz:
I have started adding the additions/alterations you asked for. From my googling, it doesn't look like Liquid has a way of checking to see if a file exists for the authors. 

As page.author is the frontmatter we are using, I would suspect that an end user would expect to see their name come up even if they didn't set up the collection for authors. Instead of not putting the byline there, how about we put the author's name in, but if their name does not exist in the collection, we don't create the name as a link?

daattali:
To check if the author exists, I think this should work:

```
{% assign author_ids = site.authors | map: "username" %}
{% if post.author and author_ids contains post.author %} ...
```

Although now that you brought up the idea of showing the author even if no page exists - I'm thinking that maybe that should always be the case, for simplicity sake. Get rid of the collection, get rid of additional files per author. Simply allow you to add a `author: name` to a post, and if that field exists, then show the authors name. No other info about the author is required. You could create a `authors.md` page that loops through all the posts and lists all the authors, similar to the `tags` page. I know it's not as feature-rich as your original idea, but it does simplify the code a lot which is a major factor, and it still achieves the main objective of allowing names. What do you think?

sschwetz:
Yeah that seems like a much better idea. I will set it up so that it will look for author: or authors: just to cover all bases in case a user thinks that they should use the plural if they have multiple authors

daattali:
Let's have only `author`, and [in the documentation ](https://github.com/daattali/beautiful-jekyll#supported-parameters) it can be highlighted that the field can be a single value or a list

daattali:
This PR seems to have more files than it needs to :)

sschwetz:
Hmm, I don't know how I managed to do that. Reverted head and forced push seem to have fixed it. I have also merged the changes in the master branch into this commit to resolve the conflict with the font-awesome change

daattali:
I've made some more changes and merging now, thanks

PR #1221, opened 2023-10-17
Set Pagination buttons to show left and right arrows instead of text newer and older

sschwetz:
This pull request updates _layout/home.html to replace the current paginator navigation buttons from words "newer" and "older" with font-awesome arrow icons. The image below shows the styling of an inactive button (left arrow) and one that has a mouse over (right arrow)

I was tossing up the idea of making the format of the icons the same as the others in the footer but decided on this format instead, as it makes them a little more obvious

image

### DIFF
--- a/_layouts/home.html
+++ b/_layouts/home.html
@@ -101,12 +101,22 @@ layout: page
 <ul class="pagination main-pager">
   {% if paginator.previous_page %}
   <li class="page-item previous">
-    <a class="page-link" href="{{ paginator.previous_page_path | absolute_url }}">&larr; Newer Posts</a>
+    <a class="page-link" href="{{ paginator.previous_page_path | absolute_url }}">
+      <span class="fa-stack fa-lg" aria-hidden="true">
+        <i class="fas fa-circle fa-stack-2x"></i>
+        <i class="fas fa-arrow-left fa-stack-1x fa-inverse" alt="Newer Posts"></i>
+      </span>
+    </a>
   </li>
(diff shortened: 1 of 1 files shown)


daattali:
https://github.com/daattali/beautiful-jekyll/discussions/1222

cyaris:
I've actually made this change manually on my Jekyll site. The arrows only I did for small windows (mobile). Would love to see a change like that implemented.

daattali:
That's an interesting idea, only removing the text on small screens. I would still like to change the right arrow from the ugly looking one that we currently have to the nicer fontawesome one.

@sschwetz what do you think about this?

cyaris:
I'd be glad to submit a pull request with my own implementation, though I'm not sure my approach should be set in stone (made two elements, one for regular and one for mobile, then two media queries to hide one when the window size was greater/less than a 700px).

daattali:
I think the clean solution would be to wrap the text in an HTML tag and use CSS to hide that tag when the screen is small.

daattali:
@sschwetz does this idea make sense to you? Changing the HTML entity arrows to fontawesome arrows, and on small screens removing the text

sschwetz:
Yes that is a much better idea than my solution. I am happy for you to run with it, whilst I work on the authors, or I will look at it once I have authors finished

daattali:
Done https://github.com/daattali/beautiful-jekyll/commit/d8e099d171962663b04fe26e0a018404edcb7e54

PR #1224, opened 2023-10-18
Add robots.txt

sschwetz:
This will add a robots.txt file to the root of the website.

As the theme generates a sitemap.xml, I am declaring it's location as per spec to assist with the site being spidered without having to go to every search engine and add yourself to their webmaster tools.

This should help with new starters getting their sites into the rankings, buy today and ill throw in a free set of steak knives.

### DIFF
--- a/robots.txt
+++ b/robots.txt
@@ -2,3 +2,6 @@
 layout: none
 ---
 Sitemap: {{ "sitemap.xml" | absolute_url }}
+
+User-agent: *
+Allow: /


daattali:
I've always assumed that for simple static content websites, where generally all pages should be crawled, a robots file doesn't have any benefits.  Based on your comment, I'm getting the sense that I was wrong... For my understanding, can you show me specific cases where the robots file is helpful even if it just includes all pages?

sschwetz:
You don't need a robots.txt file to get the site crawled. By default, all pages are allowed unless denied. However, as the theme creates a sitemap.xml file, you must tell search engines where it is. 

This is done by going to each search engine's admin pages and adding it manually or by putting it in the robots.txt file, which all crawlers will read and then follow the sitemap. [Google's Documentation](https://developers.google.com/search/docs/crawling-indexing/robots/create-robots-txt) gives more details.

daattali:
Got it. Last question - the sitemap is being created in the default standard location `site.com/sitemap.xml` - do search engines not automatically assume that the sitemap is there, does it **need** to be explicitly said in robots file?

sschwetz:
No, they have to be submitted to the search engines, or put in robots.txt

PR #1226, opened 2023-10-26
Made the footer stick to the bottom; fixes #576

### LINKED ISSUE #576: Footer stick to the bottom when contents are short or absent

ch0c0l8ra1n:
Makes the footer stick to the bottom when the content isn't enough to push the footer past the screen using CSS grids.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased version
 - BREAKING CHANGE: Allow changing the order of the social network links that appear in the footer (#1152)
 - BREAKING CHANGE: `google-scholar` social network link no longer requires the prefix `citations?user=`; if you previously set this parameter, it needs to be updated (#1189)
+- The footer of a page always sticks to the bottom, even on short pages (#576)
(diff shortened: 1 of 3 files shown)


daattali:
Thanks, this very simple solution really does seem to work! Thank you so much

daattali:
@ch0c0l8ra1n this seems to have caused a bug. Code blocks do not have horizontal scrollbars, so if there's a code block with a long line, the entire page is wider and has to scroll. See https://github.com/daattali/beautiful-jekyll/issues/1235#issuecomment-1799085996

Can you think of a good fix?

PR #1227, opened 2023-10-27
Refactor header.html

sschwetz:
When working on the header.html include file, I noticed that it was mentioned at the top that it was a mess and needed refactoring.

I have completed my take on it 😃

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -1,83 +1,52 @@
-<!-- TODO this file has become a mess, refactor it -->
-
 {% assign date_format = site.date_format | default: "%B %-d, %Y" %}
-
-{% if page.cover-img or page.title %}
 
 {% if page.cover-img %}
   <div id="header-big-imgs" data-num-img={% if page.cover-img.first %}{{ page.cover-img.size }}{% else %}1{% endif %}
     {% for bigimg in page.cover-img %}
-    {% assign imgnum = forloop.index %}
-    {% for imginfo in bigimg %}
-      {% if imginfo[0] %}
-      data-img-src-{{ imgnum }}="{{ imginfo[0] | absolute_url }}"
-      data-img-desc-{{ imgnum }}="{{ imginfo[1] }}"
-    {% else %}
-      data-img-src-{{ imgnum }}="{{ imginfo | absolute_url }}"
-    {% endif %}
-    {% endfor %}
+     {% assign imgnum = forloop.index %}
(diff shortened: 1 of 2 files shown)


daattali:
Whoo, I completely forgot about that TODO! Luckily, I just did a search and saw that there are no other TODOs :) Thanks a lot for the initiative.

This is a very welcomed change, but it does need to fix one small issue. The reason for the current code being so awkward is because back in 2014 many phones still had very small screen widths. So small that when there was a cover image, I wanted the title to be below the image instead of on top of the image. It looked like this on screens smaller than 365px wide:

image

In order to support this, I had to add that entire `no-img` div in the header HTML, and some CSS: https://github.com/daattali/beautiful-jekyll/blob/8854d96eab09b1c696bbdef8c1d1ff55cd923737/assets/css/beautifuljekyll.css#L743-L774

(Note the comment in that CSS code: "Many phones are 320 or 360px")

Your changes do not account for this case. This is how that same page looks with the new code: 

image

But, I don't think it makes sense to "fix" it, because today there are almost no phones that are that small. So all that logic can be removed. For the few people in the world who still have phones from 2014, they'll just have to suffer with the title being on top of the image even on their small screens.

daattali:
Everything related to the `no-img` div should be removed. That includes the HTML that creates it (https://github.com/daattali/beautiful-jekyll/blob/8854d96eab09b1c696bbdef8c1d1ff55cd923737/_includes/header.html#L62) and any css for it

sschwetz:
I did initially remove this but noticed that this would cause the Title and subtitle on posts to end up under the nav bar 

image

If I remove the if statement, it screws with the formatting of the cover image.

image

daattali:
I implemented this in https://github.com/daattali/beautiful-jekyll/commit/9424f6d9eaba833f756be1840b1d5b652b6d6af1
Thanks for getting this started. I think I got it to look the same as before in all cases, do you mind testing it as well?

sschwetz:
I could not see any artifacts with [#9424f6d](https://github.com/daattali/beautiful-jekyll/commit/9424f6d9eaba833f756be1840b1d5b652b6d6af1)

daattali:
What do you mean by artifacts?

sschwetz:
I could not see anything that was broken with my testing of your commit

daattali:
Ah ok I was afraid you meant that you're not able to see some "assets" :) Sounds like it's all good then, closing

PR #1234, opened 2023-11-05
Setup Required files for Favicons.

sschwetz:
These are the icons that show in favourites, on tabs, and as the application icon if you pin them to the home screen on a mobile device.

This commit allows a graphic to be uploaded to https://realfavicongenerator.net and it will convert it into the required file formats. 

XML has required fields generated by liquid tags


daattali:
I seem to be able to see the favicon even without this. By simply having a file called `favicon.ico` in the root, it seems to get picked up by the browser. I tested on windows, android, and iphone. Are you not seeing this behaviour?

daattali:
I've added favicon support with a simplified version. All you need to do is add a `favicon.ico` file to the root directory. Only `<link rel="icon" href="favicon.ico" />` gets added, because I'm not convinced of the benefit of all the other lines.

Issue #1235, opened 2023-11-07
Code blocks do not scroll anymore


dagnelies:
I don't know when exactly this bug was introduced, but using the latest version of this theme, the view is now broken on mobile because of code blocks.

![image](https://github.com/daattali/beautiful-jekyll/assets/5452653/923d3b13-20e3-49c1-97f8-6cd4ec4e0d14)

I'm pretty sure that before such code blocks had scroll bar, so that the view didn't get bigger than the screen. Dunno when or why this happened.

Here is the same using a (much) older version of the theme:

![image](https://github.com/daattali/beautiful-jekyll/assets/5452653/333f4c5b-d2cf-4fed-8814-701c44c27850)

daattali:
Thanks for reporting. Going through the list of recent changes, the one that is the most likely to have caused this is perhaps https://github.com/daattali/beautiful-jekyll/commit/ba46ef8abd1e2aea30c04b633a7fcdbb31e122b2

When I find some time I'll try to look into it. If you know how to, can you try to troubleshoot and see if that change indeed caused the issue?

daattali:
I've done a bit of testing and that commit is indeed was breaks it. It was introduced from https://github.com/daattali/beautiful-jekyll/pull/1226 - I'll see if the author of that PR has a fix

dagnelies:
Thanks for investigating and pin-pointing it to this commit 👍 
I think the issue comes from the `display: grid`, I'll take a look later on.

daattali:
Thanks, I'd still like to give @ch0c0l8ra1n a chance to add any info

daattali:
Thanks for @dagnelies in your help with https://github.com/daattali/beautiful-jekyll/pull/1236

I ended up making some HTML changes so I used your idea and implemented it in 068bf0f63cb93b566736ade443993e898dc174a1 and cf95491c367952617141dc98ab02fd58bf0f4984

daattali:
Can you please confirm that it's fixed for you?

PR #1236, opened 2023-11-08
layout fix for #1235

dagnelies:
For #1235

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -25,13 +25,12 @@ common-js:
 {% include head.html %}
 
 <body>
-  <div>
-    {% include gtm_body.html %}
+  {% include gtm_body.html %}
+  {% include nav.html %}
 
-    {% include nav.html %}
-
+  <main>
     {{ content }}
-  </div>
+  </main>
 
   {% include footer.html %}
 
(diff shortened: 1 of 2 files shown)


daattali:
Does this fix also take into account still supporting the original reason for https://github.com/daattali/beautiful-jekyll/commit/ba46ef8abd1e2aea30c04b633a7fcdbb31e122b2 ? In other words, is https://github.com/daattali/beautiful-jekyll/issues/576 still fixed as well? Or does this PR only fix the current issue?

dagnelies:
Yup. Both fixed

daattali:
Great, I'll test it later, thanks! Looking at the code changes, I'd like to change the `<main>` to a `<div>`, because there are already divs further down that have `role="main"`, and I don't think it's good practice to have `<main>` and inside of it another div with `role="main"`

dagnelies:
Sure. Do as you like. Please pay attention that the CSS rule `body > main { flex: 1 }` should only target the main area of the body, not the nav header or footer.

dagnelies:
you could even drop this outer "main" container directly and apply the `flex:1` to the main content container.

daattali:
@dagnelies Thanks for all your work on this. I just updated the layouts so that every page now has a `<main>` tag - however, it's not in the exact same hierarchical location within the DOM on each page type. Would applying `flex:1` to the main tag  would solve the problem?

dagnelies:
The CSS rule `body > main { flex: 1 }` makes the main grow to fill the empty page if necessary. As long as there is a `<main>` directly under the `<body>`, anywhere, what's after the main will appear at the bottom.

PR #1249, opened 2023-12-13
Add favicon

tae898:
This adds a small icon of your choice in the left corner of the browser tap.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Added social network links for GitLab, Bluesky (#1168, #1218)
 - Added instructions and example on how to fix image links in project sites (#1171)
 - Pagination buttons: use nicer arrows, and don't show text on small screens (#1221)
+- Add a favicon (#1249)
 
 ## v6.0.1 (2023-06-08)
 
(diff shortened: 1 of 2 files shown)


tae898:
Ah, obviously there is already an existing PR on this: https://github.com/daattali/beautiful-jekyll/pull/1234

My simple version works fine, though.

daattali:
From my experience over the last 10 years, I've made many websites using beautifuljekyll and I never had to explicitly add the favicon location to the code. I simply place a favicon in the root, and it works (https://beautifuljekyll.com/faq/#favicon). Was there some change recently that makes this no longer the case? I'm trying to understand the need for this PR

tae898:
Yeah somehow for my website, if I place `favicon.ico` in the root directory, it doesn't work. The favicon appears and disappears immediately from the web tab. I don't know why.

daattali:
I'm wondering if it's a caching issue. Do you have a beautifuljeklly site, or can you create one, that has this issue? Where you have a favicon in the root and it doesn't automatically pick it up, I want to troubleshoot it

daattali:
I've added favicon support with a simplified version. All you need to do is add a `favicon.ico` file to the root directory. Only `<link rel="icon" href="favicon.ico" />` gets added, because I'm not convinced of the benefit of all the other lines.

Issue #1192, opened 2023-08-27
Post Header image option


vakesz:
I don't know if this is an option inside the premium version, but on my side I didn't like the thumbnail design on beautiful-jekyll, and created an option to have a horizontal header image for posts.

PR: https://github.com/vakesz/vkszblog/pull/5
The code is quite jank, so probably would require some optimizations.

Preview:
![image](https://github.com/daattali/beautiful-jekyll/assets/3627611/e4822fd7-ba05-4477-91e1-981734f2a52b)

-----------------------------
Please only submit feature suggestions or bug reports if you believe something is broken.

If you need help, you can attend the [Office Hours](https://beautifuljekyll.com/officehours) (only available for [sponsors](https://beautifuljekyll.com/plans/)).

daattali:
That looks good, I'd be happy to introduce that to beautifuljekyll. The code would need to be very clean (for maintenability) and there would need to be a config parameter to specify whether the current behaviour should be used or if the thumbnail should be horizontal. The layout would need to be sure to look good on both desktops and mobile.

vakesz:
The code I linked in the Issue as a PR link already scales within mobile, and desktop. But I know width set to 1000px(which gets limited by the post width) is not a nice solution but I haven't played much with CSS so it's mostly a grey area for me. 

Maybe one addition that would be nice to have an option like so:
```
# post-preview-style: header, thumbnail, header-mobile-thumbnail-desktop, none
# header: Header bar shown on post
# thumbnail: Thumbnail shown on post
# header-mobile-thumbnail-desktop: as the name suggests
# off: no pic on post
```

daattali:
@vakesz do you want to submit a PR?

daattali:
@vakesz this issue will be closed if no response

vakesz:
Sorry for the delay. Surely if i'll get some more free time I would be happy to update it.

daattali:
Closing because it won't get done by anyone else - but feel free to re open in the future if you'd like to work on it

PR #1262, opened 2023-12-31
[WIP] Homepage styling options

vakesz:
Issue reference: https://github.com/daattali/beautiful-jekyll/issues/1192

### DIFF
--- a/_layouts/home.html
+++ b/_layouts/home.html
@@ -54,7 +54,13 @@ layout: page
         Posted on {{ post.date | date: date_format }}
       </p>
 
-      {% if thumbnail != "" %}
+      {% if site.post_header_img == wide %}
+      <div class="">
+         <a href="{{ post.url | absolute_url }}" aria-label="Header">
+            <img src="{{ thumbnail | absolute_url }}" class="post-header-image" alt="Post header">
+         </a>
+      </div>
+      {% elsif site.post_header_img == square %}
       <div class="post-image post-image-small">
         <a href="{{ post.url | absolute_url }}" aria-label="Thumbnail">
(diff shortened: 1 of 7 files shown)


_layouts/home.html
```diff
@@ -54,7 +54,13 @@ <h3 class="post-subtitle">
         Posted on {{ post.date | date: date_format }}
       </p>
 
-      {% if thumbnail != "" %}
+      {% if site.post_header_style == "wide" %}
```
daattali, line 57: Shouldn't this still be wrapped in the `if thumbnail !- ""`? Because if there's no thumbnail, then this should not be run?

_layouts/home.html
```diff
@@ -63,7 +69,7 @@ <h3 class="post-subtitle">
       {% endif %}
 
       {% unless site.feed_show_excerpt == false %}
-      {% if thumbnail != "" %}
+      {% if thumbnail != "" and site.post_header_style == "none" %}
```
daattali, line 72: Maybe where `thumbnail` gets defined above, there should be a check to see if the user is choosing to not show image preview. If the user chooses "none", then set `thumbnail` to be empty

daattali (changes requested):

daattali:
Thanks for the initiative. I made some comments

daattali:
@vakesz do you want to finish this PR?

daattali:
@vakesz ping :)

daattali:
@vakesz would you like to finish this PR?

vakesz:
Will reopen if I'll have more time. Even though all changes are relatively small I can't put time into this topic. Sorry though.

Issue #1263, opened 2024-01-01
Search Bar Placeholder And Text


TartejBrothers:
- The font family for the search bar looks out of place when compared to the rest of the website.

daattali:
You're right, it's a different base font. Feel free to make a PR!

PR #1265, opened 2024-01-04
Updated Search Bar 

TartejBrothers:
Fixing Issue #1263, updated the search bar

### DIFF
--- a/assets/css/beautifuljekyll.css
+++ b/assets/css/beautifuljekyll.css
@@ -1003,6 +1003,7 @@ pre {
   text-align: center;
   padding: 1rem;
 }
+
 #nav-search-exit {
   position: absolute;
   top: 1.5rem;
@@ -1030,6 +1031,10 @@ pre {
   outline: none;
   border: none;
   padding: 0 3rem;
+  font-family: 'Open Sans', 'Helvetica Neue', Helvetica, Arial, sans-serif;
+}
+#nav-search-input::-webkit-input-placeholder {
+  font-family: 'Open Sans', 'Helvetica Neue', Helvetica, Arial, sans-serif;
 }
 @media (max-width: 1199px) {
   #nav-search-input {
--- a/assets/css/beautifuljekyll.css
+++ b/assets/css/beautifuljekyll.css
@@ -992,6 +992,7 @@ pre {
 
 /* Search bar */
 #beautifuljekyll-search-overlay {
+  font-family: 'Open Sans', 'Helvetica Neue', Helvetica, Arial, sans-serif;
   display: none;
   z-index: 999999;
   position: fixed;
@@ -1031,7 +1032,6 @@ pre {
   outline: none;
   border: none;
   padding: 0 3rem;
-  font-family: 'Open Sans', 'Helvetica Neue', Helvetica, Arial, sans-serif;
 }
 
 @media (max-width: 1199px) {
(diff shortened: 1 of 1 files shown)


TartejBrothers (commented):
Added Font Family

daattali:
- Is the css for the placeholder necessary? I tried it without explicitly adding placeholder css, and it seems like the placeholder text still inherited the font from the input.
- What about the text of the results?

TartejBrothers:
Ok, i will update that

TartejBrothers (commented):
Removed Placeholder and added font family to result container anchor

daattali:
Am I correct in my assessment though, or was the placeholder css used for a reason?

TartejBrothers:
Yes, you are correct, its not required but I used just in case if it was inheriting font-family from somewhere else. Here in this case, thats not happening, so it will inherit input field font-family.

daattali:
I don't think this will account for the case of "No results found". I think the css can be applied only to `#beautifuljekyll-search-overlay` and that will take care of everything

TartejBrothers:
Hi, any updates on this

TartejBrothers:
@daattali Any updates on this

assets/css/beautifuljekyll.css
```diff
@@ -1003,6 +1003,7 @@ pre {
   text-align: center;
   padding: 1rem;
 }
+
```
PublisherApplications, line 1006: Thanks for adding thing.

assets/css/beautifuljekyll.css
```diff
@@ -1030,7 +1031,9 @@ pre {
   outline: none;
   border: none;
   padding: 0 3rem;
+  font-family: 'Open Sans', 'Helvetica Neue', Helvetica, Arial, sans-serif;
```
PublisherApplications, line 1034: Thanks for adding thing.

assets/css/beautifuljekyll.css
```diff
@@ -1030,7 +1031,9 @@ pre {
...
 }
+
```
PublisherApplications, line 1036: Thanks for adding thing.

assets/css/beautifuljekyll.css
```diff
@@ -1065,10 +1068,12 @@ pre {
   font-size: 1.5rem;
   max-height: calc(100vh - 6.5rem);
   overflow-y: auto;
+  font-family: 'Open Sans', 'Helvetica Neue', Helvetica, Arial, sans-serif;
```
PublisherApplications, line 1071: Thanks for adding thing.

assets/css/beautifuljekyll.css
```diff
@@ -1065,10 +1068,12 @@ pre {
...
 }
 #search-results-container a {
   color: #fff;
   text-decoration: none;
+  font-family: 'Open Sans', 'Helvetica Neue', Helvetica, Arial, sans-serif;
```
PublisherApplications, line 1076: Thanks for adding thing.

PublisherApplications (commented):
This changes are are approve

TartejBrothers:
> > I think the css can be applied only to `#beautifuljekyll-search-overlay` and that will take care of everything
> 
> This would be a simpler fix, only one line should be added

I added that in the commit.

daattali:
That's the **only** line that is required. No other lines need to change. Just one line is enough for the entire commit, no?

TartejBrothers (commented):
Removed the extra font css too

PR #1268, opened 2024-01-05
Add itch.io social network link to the commented out entries

Chadowo:
As the title implies, this PR adds a itch.io entry to the social network links along with all the others that are commented out. I actually didn't know that this template supported itch.io as social network link, since I only looked in the `_config.yml` and though the ones present there were the only ones available. So I think adding it there will make it somewhat less obscure.


daattali:
I don't know how that happened! Thanks for fixing

PR #1270, opened 2024-01-10
Move hard version constraint to Gemfile to allow gem to run on 4.x

pedantic-git:
Hi @daattali - thank you for an amazing theme!

I was using version 6.0.1 with a Jekyll 4.3.3 site and it was working fine, but I needed some of the fixes that have been made since then (for example, a206f70), so I tried to switch to using the latest master on my site and that's when I fell foul of 1758870, which forces the version to 3.x.

The gem works fine with 4.x in my limited testing, so would it be possible to restore support for 4.x? Perhaps if the constraint is just for GH Actions, it could be moved out of the gemspec and into the `Gemfile`? Here is a PR to that effect.

Thanks so much!

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,5 +2,6 @@
 
 source "https://rubygems.org"
 
-gemspec
 
+gemspec
+gem 'jekyll', '~> 3.9.3'
--- a/beautiful-jekyll-theme.gemspec
+++ b/beautiful-jekyll-theme.gemspec
@@ -17,7 +17,7 @@ Gem::Specification.new do |spec|
     "documentation_uri" => "https://github.com/daattali/beautiful-jekyll#readme"
   }
 
-  spec.add_runtime_dependency "jekyll", "~> 3.9.3"
+  spec.add_runtime_dependency "jekyll", ">= 3.9.3"
   spec.add_runtime_dependency "jekyll-paginate", "~> 1.1"
   spec.add_runtime_dependency "jekyll-sitemap", "~> 1.4"
   spec.add_runtime_dependency "kramdown-parser-gfm", "~> 1.1"


daattali:
Last I tried, being flexible caused github actions to fail. Since this theme is github-pages-first (meaning that it prioritizes working smoothly on GitHub over any other platform), I would not make any changes that negatively affect it on GH. 

You can see from the history of that file that this was needed https://github.com/daattali/beautiful-jekyll/commits/master/beautiful-jekyll-theme.gemspec

You're welcome to try making the change again and see if github actions work with a flexible jekyll version. If it works, then I'll accept the PR.

pedantic-git:
Hi @daattali - thanks for your quick reply! My PR does keep the stricter version constraint but it moves it from the `gemspec` to the `Gemfile`. Although I can't see if it passes GH Actions (the workflow is awaiting approval), I believe it will because GH Actions will install gems from the `Gemfile` (which includes the `gemspec`), whereas dependent Jekyll sites will install them from the `gemspec`. But it's just a hypothesis at the moment.

pedantic-git:
@daattali I enabled Actions on my repo and confirmed the action passes without error with this change and the override in `Gemfile`. Let me know if you have other concerns!

Gemfile
```diff
@@ -2,5 +2,6 @@
 
 source "https://rubygems.org"
 
-gemspec
 
+gemspec
+gem 'jekyll', '~> 3.9.3'
```
PublisherApplications, line 7: Thanks for adding thing.

Gemfile
```diff
@@ -2,5 +2,6 @@
 
 source "https://rubygems.org"
 
-gemspec
 
+gemspec
```
PublisherApplications, line 6: Thanks for adding thing.

Gemfile
```diff
@@ -2,5 +2,6 @@
 
 source "https://rubygems.org"
 
-gemspec
```
PublisherApplications, line 5: Thanks you for deleting this.

beautiful-jekyll-theme.gemspec
```diff
@@ -17,7 +17,7 @@ Gem::Specification.new do |spec|
     "documentation_uri" => "https://github.com/daattali/beautiful-jekyll#readme"
   }
 
-  spec.add_runtime_dependency "jekyll", "~> 3.9.3"
+  spec.add_runtime_dependency "jekyll", ">= 3.9.3"
```
PublisherApplications, line 20: Thanks for adding thing.

beautiful-jekyll-theme.gemspec
```diff
@@ -17,7 +17,7 @@ Gem::Specification.new do |spec|
     "documentation_uri" => "https://github.com/daattali/beautiful-jekyll#readme"
   }
 
-  spec.add_runtime_dependency "jekyll", "~> 3.9.3"
```
PublisherApplications, line 20: Thanks you for deleting this.

PublisherApplications (commented):
These changes are approved

pedantic-git:
I've provided a better solution to this in #1276 so I'm going to close this one now. Interested to see what you think!

PR #1276, opened 2024-01-20
Fix GitHub Actions for Jekyll 4.x

pedantic-git:
Hi @daattali - instead of papering over the issue with my previous fix in #1270, I decided to do a deep dive into why GitHub Actions was failing with the relaxed version constraint and fix that instead.

The reason is because it depended on the Docker image `jekyll/builder:3.8`, which installs Jekyll 3.8.x, but the version constraint required at least 3.9.3. This meant bundler would install the latest version that matched the constraint over the top of the 3.8.x in the image. If that was a 4.x then it would fail because 4.x depends on `sass-embedded` which is not supported by the version of Rubygems in the `jekyll/builder:3.8`. Unfortunately, [jekyll/builder is no longer maintained](https://github.com/envygeeks/jekyll-docker/issues/364) and consequently there is no image for the 3.9.x strand.

Instead, I've switched to a GitHub action based on the `ruby/setup-ruby` action which is supported by GitHub themselves, and layered on top an [appraisal](https://github.com/thoughtbot/appraisal) which allows the CI to build both the 3.9.x version and the 4.3.x version and only pass if _both_ of them build.

How does this look to you? Hopefully it ensures that GH Pages will still build but also ensures that the 4.x branch builds for those of us who are using it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,12 +4,23 @@ jobs:
   build:
     name: Build Jekyll
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        ruby_version: ['3.3']
+        jekyll_version: ['3.9.4', '4.3.3']
     steps:
       - uses: actions/checkout@v3
-      - name: Build the site in the jekyll/builder container
-        run: |
-          export JEKYLL_VERSION=3.8
-          docker run \
-          -v ${{ github.workspace }}:/srv/jekyll -v ${{ github.workspace }}/_site:/srv/jekyll/_site \
-          -e PAGES_REPO_NWO=${{ github.repository }} \
-          jekyll/builder:$JEKYLL_VERSION /bin/bash -c "chmod 777 /srv/jekyll && jekyll build --future"
+      - uses: ruby/setup-ruby@v1
+        with:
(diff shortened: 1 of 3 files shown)


pedantic-git:
Hi @daattali - I'm guessing since this has stayed open for a while you're not very comfortable with merging it. Is there more information I can provide to help with your comfort levels about this change? It's very much a "covers all bases" change that ensures the 3.9.x variants are always tested even though 4.x is allowed (unlike my previous attempt).

daattali:
I was on a 2 month vacation and while I was still working for my clients, I had very little time to dedicate to all my github projects. Back in Jan I've spent some time looking at this PR and your previous PR, I'm not opposed to it but I do need to spend more time testing it and ensuring everything about it is fine. I'm extremely conservative and careful with making changes in this project since it affects so many people and there are so many different ways to install and need to make sure it doesn't negatively affect any of them.

daattali:
Thank you so much for investigating the root cause and making this PR, I like this approach much more!

pedantic-git:
Thanks so much for following up and for looking at it more and merging it!

It makes sense that you are very careful about this gem given its impact!l. I hope my comment didn't sound like pressure rather than an offer of support but if it did I'm sorry! And I hope you had a great vacation!

Issue #1284, opened 2024-01-31
[performance] Search is creating larger pagesizes because the index is built in


monksy:
Based on the doc:
https://github.com/christian-fei/Simple-Jekyll-Search

You can generate the search text in a seperate file. I think this would help build times and reduce page sizes. 

https://github.com/daattali/beautiful-jekyll/blob/e5ababbfb57c0dbf5dae997c60271b4ccee6fed8/_includes/search.html#L12

There would be a request for the end user to create a generated page:

```md
---
layout: none
---
[
  {% for post in site.posts %}
    {
      "title"    : "{{ post.title | escape }}",
      "category" : "{{ post.category }}",
      "tags"     : "{{ post.tags | join: ', ' }}",
      "url"      : "{{ site.baseurl }}{{ post.url }}",
      "date"     : "{{ post.date }}"
    } {% unless forloop.last %},{% endunless %}
  {% endfor %}
]
```

daattali:
Fixed with e6fc968cbdd592ed73ed49eb622cbf32ed67728c

PR #1288, opened 2024-02-06
Moving search to it's own file.

monksy:
Speed up search and reduce page sizes.

https://github.com/daattali/beautiful-jekyll/issues/1284

Please note that if you are trying to update **your** website, this is the wrong place to do so. Please carefully follow the Beautiful Jekyll instructions (found at https://github.com/daattali/beautiful-jekyll#readme) and make sure you submit changes to **your** version of the project.

If your intention is to submit a Pull Request, please describe what your pull request achieves.

Thank you!

### DIFF
--- a/_includes/search.html
+++ b/_includes/search.html
@@ -1,41 +1,18 @@
 {% if site.post_search %}
 
 <div id="beautifuljekyll-search-overlay">
+    <div id="nav-search-exit" title="Exit search">✕</div>
+    <input type="text" id="nav-search-input" placeholder="Search">
+    <ul id="search-results-container"></ul>
 
-  <div id="nav-search-exit" title="Exit search">✕</div>
-  <input type="text" id="nav-search-input" placeholder="Search">
-  <ul id="search-results-container"></ul>
-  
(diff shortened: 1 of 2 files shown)


monksy:
image

monksy:
@daattali

daattali:
Thanks for the suggestion! I've been thinking about this for a while and while I do really agree with the idea, there's one problem that makes it difficult for me to accept this PR. 
Anyone who uses this theme as a `remote_them` or a ruby gem will not have access to the json file you created, and they'll be required to manually add that file. There are currently only a handful of files that behave this way, and I try to stay away from that as much as possible, because each one of these cases adds more friction. Do you think there's a way to achieve the same goal but without creating a file that users will have to create manually?

daattali:
@monksy ping :)

ccorsi:
@daattali I agree with you that this is a good idea but the implementation requires a better solution.  I think that the generated json file should be placed within the assets directory, specifically within a data subdirectory of assets.  Since your theme includes this as part of your distribution then this solution will work when using the gem or remote_theme feature.

daattali:
Implemented in e6fc968cbdd592ed73ed49eb622cbf32ed67728c

Thanks @ccorsi for your help

PR #1290, opened 2024-02-21
Relax version constraint for kramdown

pedantic-git:
Hi @daattali 

Versions of the `github-pages` gem after 229 depend on `kramdown` version 2.4.x. This PR relaxes the version constraint to ensure people can update to the latest version of `github-pages`. It should still work exactly the same with earlier versions too.


daattali:
Thank you!

PR #1299, opened 2024-03-21
Added Whatsapp to social network

SwissOS:
Adding Whatsapp to social networks. Especially useful on mobile devices to launch whatsapp directly with the correct phone number.


daattali:
That's an excellent addition, I can't believe we didn't have whatsapp until now!

PR #1311, opened 2024-04-14
add configurable rotate time for header images

smooshy:
Adding the ability to define a different timeout for the banner image rotation. Defaults to 6000 like before, but allows users of the theme to configure a different value.


daattali:
@smooshy thanks for the initiative. In theory this is a good idea, but I prefer not to add an additional parameter. I aim to keep this theme simple, and that requires walking a fine line between more features = more customizable and less features = simpler. I sometimes have to turn down configuration options to avoid the config file from becoming extremely long and unmanageable (which it arguably already is!)

Issue #1322, opened 2024-05-27
Post preview shows markdown


thejcannon:
👋 I'm trying out Beautiful Jekyll for a site and noticed that my post, which starts with a markdown code block is rendered quite literally in the home pages summary (E.g. the summary starts with "\`\`\`text")

(Also love it so far, thanks for making this ❤️)

daattali:
This is unfortunately working "as intended". It's not really an issue specific to this theme, it's a more general jekyll concern. I'm going to close this issue since it's not a bug with my theme, but I'll try to walk you through what's happening, so that you can try to think of a custom solution or take this info and ask in a jekyll forum about work arounds:

Jekyll has the concept of a "post excerpt" as described here https://jekyllrb.com/docs/posts/#post-excerpts and as you can see, by default it takes the **first paragraph of content**. Here if the code in beautiful jekyll that takes this excerpt and shows it on the home page: https://github.com/daattali/beautiful-jekyll/blob/b57d3d4a4abd3027edc8c39e2749a57b53f9a5e2/_layouts/home.html#L75-L76

So what's happening is that because you have a blank line inside that (poem?), jekyll treats that as a new paragraph. So it takes all the content up until that blank line, which is:

````
```text
It is a period of civil war.
Rebel collectors, striking
from a hidden basement,
have won their first victory
against the evil Garbage Empire.
````

And this it the post excerpt (if you use the `excerpt_length` parameter that beautifuljekyll uses, then it'll truncate the number of words). When you take the text above, jekyll does not know that this is markdown because it doesn't see the closing backticks. So it thinks the opening ticks are part of the text.

To drive this point further, I think (I haven't tested this) that if you remove blank lines from the poem, or if you ended the markdown syntax after the first paragraph, then the post excerpt on the home page will look correct. 

You can also explicitly define what the excerpt should be, by adding a `excerpt:` YAML field in the post. On the jekyll documentation page that I linked above, you may have noticed that they talk about a `excerpt_separator`, perhaps you can use that so that the blank line doesn't end the excerpt.

I hope this was useful.

thejcannon:
Oh wow that's a very thorough response! That all makes perfect sense, I'll play around with some settings, thanks!

PR #1323, opened 2024-05-28
Add `excerpt` to README

thejcannon:
https://github.com/daattali/beautiful-jekyll/issues/1322 mentioned this setting, but I didn't see it on the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,7 @@ These are parameters that you may not use often, but can come in handy sometimes
 Parameter   | Description
 ----------- | -----------
 author      | Specify the author of a blog post (useful if a website has multiple authors).
+excerpt     | Specify the excerpt to show on the home page.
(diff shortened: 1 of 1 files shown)


thejcannon:
That's fair. Since the README says "below is a list of all parameters" it makes me think the wording should be adjusted and a link to the base Jekyll metadata be given.

Otherwise:
1. It's not obvious there's additional fields beyond these
2. It's also not obvious where to find documentation of the fields.

What do you think?

daattali:
I really want to stay away from mentioning *this specific part of jekyll (the additional YAML parameters)*, because there are so many other features of jekyll that can also be mentioned. You found this feature useful this week, but there's nothing inherently more interesting about this than many other parts of their documentation. Because of that, I prefer to just link to their documentation. Because I want to keep the README as simple as possible and not to overwhelm beginners, I'll add it to the **Next Steps** section.

PR #1344, opened 2024-07-09
index page updated after adding JPG to image folder

rakaspace:
index page updated after adding JPG to image folder


rakaspace:
nothing to comment

PR #1389, opened 2024-10-29
Add sitemap.xml

kzu:
This helps with SEO. We index all content by default (pages and posts),
unless explicitly opted-out via the `sitemap: false` front-matter.

We also exclude all static assets by default, since it's typically
made up of data, js, css and images which are typically not indexed either.

### DIFF
--- a/404.html
+++ b/404.html
@@ -2,6 +2,7 @@
 layout: default
 title: 404 - Page not found
 permalink: /404.html
+sitemap: false
 ---
 
 <div class="text-center">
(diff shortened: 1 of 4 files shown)


daattali:
I'm using the `jekyll-sitemap` plugin, which silently adds a sitemap already. Is that not sufficient? (disclaimer: I haven't looked at or thought about sitemaps in many years)

kzu:
good pointer! I didn't know about it (pretty n00b to the jekyll ecosystem too)

PR #1390, opened 2024-10-29
Add missing windows extensions

kzu:
When running on Windows, wdm is required in order to properly do incremental updates and regeneration.

The tzinfo-data dependency is also reported missing (under Windows at least, with jekyll 3.9.3).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,14 +2,15 @@
 
 source "https://rubygems.org"
 
-gem "wdm", "~> 0.1.1", :install_if => Gem.win_platform?
-
 # Windows and JRuby does not include zoneinfo files, so bundle the tzinfo-data gem
 # and associated library.
 platforms :mingw, :x64_mingw, :mswin, :jruby do
-    gem "tzinfo", ">= 1", "< 3"
-    gem "tzinfo-data"
-  end
+  gem "tzinfo", ">= 1", "< 3"
+  gem "tzinfo-data"
+end
+
+# Performance-booster for watching directories on Windows
+gem "wdm", "~> 0.1", :platforms => [:mingw, :x64_mingw, :mswin]
   
 gemspec
 
--- a/Gemfile
+++ b/Gemfile
@@ -2,5 +2,14 @@
 
 source "https://rubygems.org"
 
+gem "wdm", "~> 0.1.1", :install_if => Gem.win_platform?
+
+# Windows and JRuby does not include zoneinfo files, so bundle the tzinfo-data gem
+# and associated library.
+platforms :mingw, :x64_mingw, :mswin, :jruby do
+    gem "tzinfo", ">= 1", "< 3"
+    gem "tzinfo-data"
+  end
+  
 gemspec
 


daattali:
I had no idea Windows even supports jekyll development! I'm a windows user but always assume anything I have to do has to happen on WSL on a remote server...

After looking at the docs, I see they recommend using the line `~> 0.1.1", :install_if => Gem.win_platform?` , is there a reason your is a bit different? And the timezone gem I'd rather also only add on Windows since it doesn't seem to be required by other OS

daattali:
I prefer not to modify sensitive files if we don't fully understand the impact.  Modifying html/css/js files is easy, but anything to do with ruby I try to be very cautious about. If you test this on Unix and verify that it works , let me know.

kzu:
Fixed with recommended suggestions! Run install/exec on WSL2 Ubuntu and it worked great too.

beautiful-jekyll-theme.gemspec
```diff
@@ -23,6 +23,8 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency "kramdown-parser-gfm", "~> 1.1"
   spec.add_runtime_dependency "kramdown", "~> 2.3"
   spec.add_runtime_dependency "webrick", "~> 1.8"
+  # This seems to only be required on Windows
+  spec.add_runtime_dependency "tzinfo-data", "~> 1.2023.3" if Gem.win_platform?
```
ReenigneArcher, line 27: This doesn't seem to match the suggestion in the docs. https://jekyllrb.com/docs/installation/windows/#time-zone-management

1. there are more platforms that tzinfo-data doesn't work on.
2. tzinfo should probably be defined as well?

kzu: Adding this alone makes it work for Windows. Not sure about other OSes. I guess those should be separate PRs validated by users who care for those?

ReenigneArcher: @kzu, I think something went wrong in your latest force push? It's showing a lot more files changed than before.

daattali:
I'm not sure why there are so many files changed in this PR

ReenigneArcher:
> I'm not sure why there are so many files changed in this PR

I think something was accidentally included in the last force push.

kzu:
Yep, fixing it now, sorry!

daattali:
The indentation seems to be off (there's extra spaces in a few lines).

I'm not sure if these lines would also need to be included in the gemspec file, or does it make sense to only include them in the Gemfile? @ReenigneArcher do you have insight on this?

ReenigneArcher:
I don't have a ton of experience with ruby other than a little with this project and one other homebrew spec, but I found this blog post on the topic. https://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/

That seems to suggest the dependencies should be defined in the gemspec file versus the Gemfile.

Additionally with the gemspec file they can be properly specified as dev dependencies. I guess the syntax will be slightly different though.

```gemspec
spec.add_runtime_dependency "wdm", "~> 0.1.1", install_if: Gem.win_platform?

# Windows and JRuby does not include zoneinfo files, so bundle the tzinfo-data gem and associated library.
spec.add_runtime_dependency "tzinfo", ">= 1", "< 3", platforms: [:mingw, :x64_mingw, :mswin, :jruby]
spec.add_runtime_dependency "tzinfo-data", platforms: [:mingw, :x64_mingw, :mswin, :jruby]
```

daattali:
I read that as well, which is what made me question where this should be added. I'm going to install windowsruby myself and try this out and experiment. But I'll also post on a jekyll forum to get advice from people who have much more experience. I'm going to refrain from accepting this PR until know for sure that it's correct.

daattali:
@kzu can you clarify when exactly you ran into problems with Windows?

Did you see the installation instructions described [here](https://beautifuljekyll.com/getstarted/)? Can you confirm whether Method 2 works for you - both option 1 and option 2?

daattali:
I've spent a lot of time researching this, and I **think** the issue you're having is only when you clone this repository and try to run a jekyll site using the direct files in this repo, is that correct? If you use the theme using using `jekyll new` or as a gem, there should be no issues on windows, is that correct?

daattali:
When using `jekyll new`, a Gemfile is created automatically. This Gemfile already has these tzinfo lines in it. It looks like even going back as far as jekyll 3.9.0 [this was the case](https://github.com/jekyll/jekyll/blob/v3.9.0/lib/jekyll/commands/new.rb#L92-L95). Therefore, I believe that only the local Gemfile needs to be modified, and it would only solve the case of someone trying to run jekyll by directly cloning this repo. (This is not a common usecase, but it is nice to support it).

daattali:
Before I finalize this, I asked on [Jekyll forum](https://talk.jekyllrb.com/t/tzinfo-for-windows-should-it-be-in-gemfile-or-gemspec/9523) and on [SO](https://stackoverflow.com/questions/79173271/authoring-jekyll-theme-should-tzinfo-gem-be-included-in-gemfile-or-gemspec/79175540#79175540), and I want to give it a few days to see if someone thinks my reasoning is wrong.

kzu:
Yes, I think this is mostly for the case of someone contributing to this repo or forking to make changes (as I did). :)

PR #1402, opened 2024-12-01
Bluesky icon in font awesome

pblocz:
Changed the social icon from a square to Bluesky's icon in font awesome

This is the screenshot of running this change locally:
image


daattali:
Thanks!

PR #1407, opened 2024-12-20
fix(giscus): add missing attributes and allow local theme

ReenigneArcher:
This PR adds some missing giscus attributes. I assume these were not available when giscus was added originally.

Additionally, there is a logic change that allows a local css file for the giscus theme. I tried to test this locally and it requires CORS to be disabled, I am not sure how this would work on a GitHub hosted site, but in any event it allows to specify a theme like `/assets/css/giscus.css` or any local path that starts with `/`. Normally giscus will try to get any non full url theme from their own server/api.

https://giscus.app/

Here is a screenshot of evidence of the a custom theme css working for giscus.
image


ahmadwardat (commented):
Yes

_includes/giscus-comment.html
```diff
@@ -6,9 +6,19 @@
...
+            {% assign theme = theme | absolute_url %}
+        {% endif %}
+        data-theme="{{ theme }}"
+        data-lang="{{ site.giscus.lang }}"
+        {% if site.giscus.lazy-loading == true %}
+            data-loading="lazy"
+        {% endif %}
         crossorigin="anonymous"
         async>
 </script>
```
ahmadwardat, line 24: Yes

Finsh 
@daattali 
```suggestion
 </script>_[]()@mangar - - [ ] - _
```

PR #1423, opened 2025-01-18
Update Youtube link

tsolloway:
Added the @ for YouTube to the links.html file.  For whatever reason Jekyll doesn't like it when you add it to the config. Maybe YouTube changed its formatting.


daattali:
@tsolloway when you use any non alphanumeric characters as a value in the config file, it needs to be wrapped in quotes. 

```
youtube: "@username"
```

rather than

```
youtube: @username
```

PR #1438, opened 2025-01-27
add charset attribute to script tag for simple-jekyll-search

chanwoo040531:
This PR adds the `charset="utf-8"` attribute to the `<script>` tag for `simple-jekyll-search`. This update ensures proper support for multilingual search.


daattali:
I couldn't find any examples of other projects using this script line. Can you show me a usecase where not using this charset causes issues so that I can understand it?

PR #1439, opened 2025-01-27
Use subtitles in search, feed and tags

organom:
Subtitles are not being taken into consideration for feed, search and on the tags page. This PR fixes that

Please let me know if some other important place is missing and i can add it

### DIFF
--- a/assets/data/searchcorpus.json
+++ b/assets/data/searchcorpus.json
@@ -1,23 +1,23 @@
 ---
 layout: null
 ---
-[ 
-{% for post in site.posts %} 
-{ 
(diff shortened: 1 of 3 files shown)


daattali:
Thanks for the initiative! However, not including the subtitle on the search results was a deliberate choice. The subtitle is already included in the RSS feed, inside the description tag (since there isn't a subtitle tag, and I thought a title should be kept short and concise). 

I can see a case being made to add the subtitle to the tags page. I would support that, but it shouldn't be part of the title, it will need to be on a new line and smaller. I'm closing this PR, but free to send a PR for the tags page alone!

PR #1441, opened 2025-01-31
Add margin-bottom to social media links

timangus:
The social media links at the bottom of the page lack a margin-bottom meaning that when there are many of them and they spill onto a second line, there is no space between the rows, and this looks ugly. This change adds a suitable margin and also reduces the footer's overall margin-bottom to match the increase in item margin-bottom.


ReenigneArcher:
I've noticed this as well when viewing on mobile.

daattali:
Thank you @timangus for noticing this and fixing it! I can't believe I never noticed it myself. Apologies for taking so long to react.

Issue #1440, opened 2025-01-31
#social-share-section anchor color always #fff


timangus:
In beautifuljekyll.css the anchor color for #social-share-section is [unconditionally set to white](https://github.com/daattali/beautiful-jekyll/blob/master/assets/css/beautifuljekyll.css#L947). Added in c07c3ea3, it's not really obvious what its purpose is. For almost all of the social share links it doesn't matter as their color is white anyway, but for those that aren't (e.g. reddit), it overrides the correct color with white, due to specificity rules. I think it should just be removed altogether, unless there is a good reason for it?

daattali:
I think that within that small section, there should be colour consistency, to not make it look too messy. That's at least my thinking. I just tried putting a red reddit icon in one of them, and it looked out of place to me. Do you think it looks better?

timangus:
That's fine in principle, but the problem is the background of the Reddit icon is light grey, so with the foreground colour forced to white, it looks bad and is hard to see:

![Image](https://github.com/user-attachments/assets/4a0a9c9b-1937-4047-a138-3288c433d76f)

(This is with an unmodified beautiful-jekyll other than toggling `reddit: true` in `share-links-active`.)

Better just to trust the colours from bootstrap-social I think, or if you must override the colours then override both the foreground *and* the background colours.

daattali:
You're right! I didn't even remember that reddit was supported -- I was going to tell you that it's a good idea to include reddit. It seems like it never made it into the config file, which is why I didn't remember about it. I'll remove this css line and add reddit to the config, thanks!

PR #1444, opened 2025-01-31
Add Bluesky social share link option

timangus:
This change adds a Bluesky social share link option. Note I've added CSS to bootstrap-social for Bluesky, using appropriate colors derived from the Bluesky guidelines. I'm not sure if the local copy of bootstrap-social is the correct place to be adding these really, but given upstream hasn't see any updates in years (and obviously predates Bluesky), it seems unlikely that any future conflicts are going to be introduced here.


daattali:
I appreciate the effort and your initiative, but I prefer not to add Bluesky to this very exclusive list. Bluesky is a much much less popular social network, it looks like it has less than 30M users (that's actually less than what I thought!). So in the interest of simplicity, I prefer not to add it. If/when it becomes a big player, I will happily revive this PR :) I hope you understand!

timangus:
Fair enough, though I will say from my perspective it seems like it already is a big player. Maybe it depends what country you're in, but for example in the UK I regularly see TV programmes advertising for audience engagement on Bluesky alongside the usual FB/Instagram/Twitter networks; I appreciate this is anecdotal to some extent. On the other hand you do see users with a large following on both Twitter and Bluesky claim that engagement levels are greater on the latter, the theory being that while Twitter has an objectively larger user base, a significant fraction of this are inactive accounts and/or bots. We've also got heads of state such as the Prime Minister of Canada using it, so it's not like it's some niche thing.

Anyway, no worries, I understand, just seems like it's something that people would want sooner or later, if not already.

(I also discovered issue #1440 while adding this btw; I think that probably needs attention.)

daattali:
I did a bit more research, it does look like the UK has the highest users per capita, which would explain why you see that! But it's still a very insignificant number of users relative to social media (every list of soical media platforms I looked at shows that Bluesky is way way down the list in # of users). But it also looks like it has explosive growth in the last half year, so maybe if it keeps going, it'll become globally relevant. But for now, while the absolute number is still very low, I'll stand by my decision :) But I'll reopen this issue so that I'll remember to revisit this later.

PR #1469, opened 2025-05-02
fix: import logger and bigdecimal gems that will no longer be included in ruby 3.5.0

fsferrara:
logger and bigdecimal will no longer be part of the default gems starting from Ruby 3.5.0
I am no ruby expert, but I think them have to be explicitly declared.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -10,7 +10,7 @@
 ############################
 
 # Name of website
-title: fsferrara
+title: My website
 
 # Your name to show in the footer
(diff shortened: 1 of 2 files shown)


_config.yml
```diff
@@ -10,7 +10,7 @@
 ############################
 
 # Name of website
-title: My website
+title: fsferrara
```
fsferrara:
```suggestion
title: My website
```

